PR #5, opened 2016-03-25
Smaller text in controller (machine list) removed tooltip

tomevoll:


raoulvdberge:
Thanks!

tomevoll:
i just noticed there is a issue when you have the controller gui open, not related to this text change, but it seems the list of machines get changed or updated causing crash.

Gives a index out of bounds or null pointer is random, is related to 
                IBlockState machineState = machine.getWorld().getBlockState(machine.getPos());

My guess is a scan just happend and the list was modified and the gui read that machine from the list at the exact point it was null.

tomevoll:
i added a function in the gui, you may wanna move it to a more global place, it allows you to recalculate the position of things when you apply a scale so they get rendered in the correct position :)

tomevoll:
sometimes the crash happends right away, other times you need to keep the gui open for a long time

raoulvdberge:
I'll investigate that crash, I think I had it too.

tomevoll:
is a small bug when turning the controller off with redstone, the grid is still on, and is listed in machines

tomevoll:
and power usage when it says 24 rf/t seems to be 24 rf a seccond, having the system active it draws around 1 rf/t from the generator

tomevoll:
maybe have a global var that stores energy usage then recheck the usage every sec as you do, but draw the energy every tick

tomevoll:
my mistake, you do have a global already, but you only draw energy 1 time every 20 ticks :)

PR #6, opened 2016-03-25
Fixes RF draw incorrect

tomevoll:
sorry for the clean commits in it lol, tried to match the spaces to minimize changes, for some reason it refuses to pull into that fork, even though it says it did it lol

### DIFF
--- a/src/main/java/refinedstorage/gui/GuiController.java
+++ b/src/main/java/refinedstorage/gui/GuiController.java
@@ -49,15 +49,10 @@ public class GuiController extends GuiBase {
         scrollbar.draw(this);
     }
 
-<<<<<<< HEAD
+
     private int calculateOffsetOnScale(int pos ,float scale) {
     	float multiplier = (pos / scale);
     	return (int)multiplier;
-=======
-    private int calculateOffsetOnScale(int pos, float scale) {
-        float multiplier = (pos / scale);
-        return (int) multiplier;
->>>>>>> 002b5c0dda19a3eb8bfddf2cc6182cbdc1b9bbee
     }
 
     @Override
@@ -82,11 +77,8 @@ public class GuiController extends GuiBase {
 
                 ItemStack machineStack = new ItemStack(machineBlock, 1, machineBlock.getMetaFromState(machineState));
 
-<<<<<<< HEAD
-                drawItem(x, y+5, machineStack);
-=======
                 drawItem(x, y + 5, machineStack);
->>>>>>> 002b5c0dda19a3eb8bfddf2cc6182cbdc1b9bbee
+
                 GlStateManager.pushMatrix();
                 float scale = 0.5f;
 
--- a/src/main/java/refinedstorage/gui/GuiController.java
+++ b/src/main/java/refinedstorage/gui/GuiController.java
@@ -49,7 +49,6 @@ public class GuiController extends GuiBase {
         scrollbar.draw(this);
     }
 
-
     private int calculateOffsetOnScale(int pos ,float scale) {
     	float multiplier = (pos / scale);
     	return (int)multiplier;
--- a/src/main/java/refinedstorage/gui/GuiController.java
+++ b/src/main/java/refinedstorage/gui/GuiController.java
@@ -49,7 +49,7 @@ public class GuiController extends GuiBase {
         scrollbar.draw(this);
     }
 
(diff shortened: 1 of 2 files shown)


tomevoll:
about the bug i mentioned with redstone, that is if redstone is set to ignore. not sure if maybe it is disabling neightbor blocks, works when its set to use redstone, so that may not be a bug at all

raoulvdberge:
I'd rather not make it draw RF every tick, but every 20 ticks?

tomevoll:
its reporting 0 rf/t when turned off with redstone, but is drawing 10 rf/t, not sure if i did that changing what i just did, ill look at it

raoulvdberge:
It can never draw 10 RF/t as the extractEnergy call is only called when isActive().

raoulvdberge:
Oh no, it actually can, it's outside the if block, my mistake. :P Fixed

tomevoll:
ya lol, i forgott to add that check when i moved it

tomevoll:
you need the powerdraw outside the % 20 block, or it will draw every seccond not every tick :)

raoulvdberge:
Fixed, thanks

raoulvdberge:
Also there seems to be an issue with importers importing from ISidedInventory's slots they can't import from. (For example: importing at bottom of furnace sucking up items from input slot)

tomevoll:
ya, need to check isidedinventory.getSlotsForFace(side); returns a int array of all slots available from the chosen side, using only getsizeInventory will ignore if it the slot is acually accessible from that side.

tomevoll:
i can take  a look at the extractor

tomevoll:
so canImport checks the filter as needed, so basicly the rest there would need a rewrite to support sided inventory properly, i should be able to do that pretty quickly :)

tomevoll:
oh wait, i didnt look close enough, dont need alot of changes at all

tomevoll:
hmm, i think the reason it pulls from the input slot is, the input is available from the side

raoulvdberge:
So how does the hopper do it when putting a hopper *under* the furnace? Shouldn't it use that method too, and why wouldn't it take from the input slot too then......

tomevoll:
the output slot is only avalible on the bottom, the input slot for item and fuel is the side, and input for item at top

tomevoll:
private static final int[] slotsTop = new int[] {0};
    private static final int[] slotsBottom = new int[] {2, 1};
    private static final int[] slotsSides = new int[] {1};

tomevoll:
ops,
Input at top
Fuel and input side
output bottom

tomevoll:
omg i mixed it up again lol
0  is Slot for Item
1 is Fuel
2 is Output

raoulvdberge:
hmm, but when having putting a importer ON a furnace, where the importer points at the furnace upwards, it takes from the input slot...

tomevoll:
hmm ok, thats weird, ill test it

tomevoll:
ya cause your not checking getSlotsForFace(side); so your pulling it from 0, canExtract wont report false, your not supose to ask it since that slot is not listed as available from that side

tomevoll:
hmm [https://github.com/tomevoll/refinedstorage/commit/acee602eeb6c807e95ec53b0b8a367f5960b4169](url)

thats one way of doing it, but currentSlot should be using only the available slots instead of the complete size of the inventory.

tomevoll:
ops dead link

Do you need to wait a few ticks between checking slots, isnt it better to just wait some ticks between each transfers

tomevoll:
[example code](https://github.com/tomevoll/refinedstorage/commit/acee602eeb6c807e95ec53b0b8a367f5960b4169)

raoulvdberge:
Thanks for the example code, I'll be doing a fix...

tomevoll:
item pulling inserting is a mess really, worst part you do in a mod, so many things can go wrong :)

tomevoll:
got plans for auto crafting ?

tomevoll:
one thing that would be nice is if sneaking while placing a importer,  external storage or importer, reverse the direction, so i can as an example go behind a furnace sneak click when placing the export and it will face the furnace not me

raoulvdberge:
Good idea, can you make an issue for it on Curse so I remember?

tomevoll:
sure

raoulvdberge:
Autocrafting is kinda meh to implement, but I'll see... :)

PR #9, opened 2016-03-28
NullPointer crash in TileMachine 

tomevoll:
null pointer on worldObj


raoulvdberge:
How can worldObj ever be null?

PR #10, opened 2016-03-28
Show true itemcount in tooltip

tomevoll:


raoulvdberge:
Why would we want this? Isn't the count display on the item itself already?

raoulvdberge:
Instead of adding the amount to the tooltip, I've made it so that if you press shift while hovering, the real amount is displayed instead of the formatted one.

f88889765f26a02358abaf5b3b97bfdd5675d4c5

PR #12, opened 2016-03-28
Fixes index out of bounds and prevent list from being modified by other threads on the client while rendering

tomevoll:
Think instead of all these null checks, maybe not set that list from the network packet as that is another thread, and instead set a temporary list that the gui forground then pull into the main list before rendering.

### DIFF
--- a/src/main/java/refinedstorage/gui/GuiGrid.java
+++ b/src/main/java/refinedstorage/gui/GuiGrid.java
@@ -133,7 +133,9 @@ public class GuiGrid extends GuiBase {
 
         for (int i = 0; i < 9 * getVisibleRows(); ++i) {
             if (slot < items.size()) {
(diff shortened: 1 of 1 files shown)


tomevoll:
no crashes after syncing the tile (makes other threads wait until your done with it)

Its only client side so should not cause any issues, when a packet arrives it waits for the tile to get unlocked.

raoulvdberge:
Wouldn't we need to add this in a bunch of other places too then? Also, I'm using an iterator in getItems(), so, nothing should be modified..

raoulvdberge:
Maybe it would be better if you added the synchronized controller statement in getItems()?

tomevoll:
ya could be, should not really matter though, its client only so the tile isnt really doing anything exept when the network packet arrives

PR #13, opened 2016-03-28
Correct RF usage, Use energy every tick not every sec

tomevoll:
Moved Energy usage outside the ticks % 20 block, amount to draw is updated every sec as before


raoulvdberge:
Dang, I thought I fixed this. Apparently not. Thanks.

PR #25, opened 2016-04-05
Store Energy in item when ccontrolelr is broken

tomevoll:
```
modified:   src/main/java/refinedstorage/block/BlockController.java
modified:   src/main/java/refinedstorage/tile/TileController.java
```

### DIFF
--- a/src/main/java/refinedstorage/block/BlockController.java
+++ b/src/main/java/refinedstorage/block/BlockController.java
@@ -14,6 +14,7 @@ import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.init.Enchantments;
 import net.minecraft.item.Item;
+import net.minecraft.item.ItemBlock;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.stats.StatList;
@@ -25,8 +26,10 @@ import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.world.IBlockAccess;
 import net.minecraft.world.World;
 import refinedstorage.RefinedStorage;
+import refinedstorage.RefinedStorageBlocks;
 import refinedstorage.RefinedStorageGui;
 import refinedstorage.item.ItemBlockBase;
+import refinedstorage.item.ItemBlockController;
 import refinedstorage.tile.TileController;
 
 import java.util.ArrayList;
@@ -44,7 +47,7 @@ public class BlockController extends BlockBase {
     @Override
     public void getSubBlocks(Item item, CreativeTabs tab, List subItems) {
         for (int i = 0; i <= 1; i++) {
-            subItems.add(new ItemStack(item, 1, i));
+            subItems.add(ItemBlockController.initNBT(new ItemStack(item, 1, i)));
         }
     }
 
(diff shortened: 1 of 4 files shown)


tomevoll:
this also store any redstone mode set on the controller

tomevoll:
im not sure if this work with silk harvest, might wanna try that and add the item generation there to, i can look at that a litle later, gotta take a break atm

tomevoll:
setEnergyStored can be removed from the tile again, since i switched to using NBT directly instead, found that to be beter since any values you write to nbt will be restored then, just make sure to not use super.writetonbt or read in there, as that will throw an error since the tileentity position is written then.

tomevoll:
lol hopefully the code works better then the title, not sure what world i was in when i typed that :P

raoulvdberge:
I think it is a better idea to just make an ItemBlock, it would be much cleaner.

See https://github.com/raoulvdberge/refinedstorage/blob/mc1.9/src/main/java/refinedstorage/block/BlockStorage.java

tomevoll:
hmm, getDroppedItems, are you sure that works? i tried that on controller and getting the tile there its always null, its already invalid at harvestBlock

raoulvdberge:
Yeah, that code works. You *need* the `removedByPlayer` and `harvestBlock` stuff too.

tomevoll:
harvest block i did, but i tried getDrops and i was unable to get the tile there witch is why i added a modified harvestBlock function, i will look at it again, world.getTile...  always gave me null back

raoulvdberge:
Did you add these 2 functions?
```java
    @Override
    public boolean removedByPlayer(IBlockState state, World world, BlockPos pos, EntityPlayer player, boolean willHarvest) {
        if (willHarvest) {
            return true;
        }

        return super.removedByPlayer(state, world, pos, player, willHarvest);
    }

    @Override
    public void harvestBlock(World world, EntityPlayer player, BlockPos pos, IBlockState state, TileEntity tile, ItemStack stack) {
        super.harvestBlock(world, player, pos, state, tile, stack);

        world.setBlockToAir(pos);
    }
```

tomevoll:
not removedByPlayer, im redoing it now with itemblock

tomevoll:
you are right that did work :)

tomevoll:
hmm, seems im missing the correct state as its not rendering the power level on the item

raoulvdberge:
I can do the item rendering stuff...

raoulvdberge:
Why are you checking if energyStored > 0?

`if (energyStored > 0) this.energy.setEnergyStored(energyStored);`

raoulvdberge:
In the `initNBT` method you forgot to set the itemstack's energy to 0...

tomevoll:
> 

Why are you checking if energyStored > 0?

if (energyStored > 0) this.energy.setEnergyStored(energyStored);


in case a negative value is read from NBT fro some unknown reason, but that function can be removed, i just forgott to remove it.

tomevoll:
dont need to set it to 0 in init unless you want it to show 0/....  when no energy is stored in it.

tomevoll:
should be better no need to do the same code 2 times

raoulvdberge:
I'll merge this if you change following things:
- Remove writeItemToNBT and readItemFromNBT, you can just call readNBT and writeNBT directly
- The energy should show up by default in the tooltip, even if it's not placed yet, so make initNBT default to 0, or, if it's a creative controller to ENERGY_CAPACITY

tomevoll:
ok

raoulvdberge:
Thanks!

tomevoll:
change: 
EnumControllerType type = EnumControllerType.values()[stack.getMetadata()];

TO:

EnumControllerType type = stack.getMetadata() == 1 ? EnumControllerType.CREATIVE : EnumControllerType.NORMAL;

in init and addinfo, some reason it crashes on values()

PR #32, opened 2016-04-25
Crafting storage + monitor textures

msrEagle:


raoulvdberge:
I think you are on an old version of the code, for example the crafting block is actually named "16k_crafting_cpu.png" in master.

raoulvdberge:
I will merge this and rename manually.

GustoniaEagle:
Na I just made those textures a week ago, only just pushed.

PR #100, opened 2016-06-10
Gradle wrapper should not be excluded from source files.

ThatSava:
It is needed by other users to setup a dev envoriment


raoulvdberge:
Thanks

Issue #124, opened 2016-06-19
Crash when loading world.


BootyToast:
I'm not sure what is happening, but after getting into Refined Storage and using it for a while I found I wasn't able to load my world the next time I joined it. Looking through the crash report it looked to be something with either Refined Storage or, for some reason, FastLeafDecay.

Crash Report: http://pastebin.com/nZRKP7Xx

BootyToast:
After uninstalling FastLeafDecay, I was able to load the world. I have a tree farm right next to my Refined Storage system, so I think there's some weird issue with there.

raoulvdberge:
Fixed in the mentioned PR, probably. If it still occurs, let me know.

PR #127, opened 2016-06-19
Fixed grid searching controller over null blocks

davenonymous:
When placing a cable above a door and triggering a searchController
call from a connected grid (e.g. by loading the chunk), then onConnected
would try to notify neighbors of a state change with block set to null.
This caused a NPE down the line.

To reproduce build this and reload the world:

Probably the same as #124


raoulvdberge:
Thanks

PR #159, opened 2016-07-05
Allow crafting upgrades to be used in the Interface to craft exported…

### LINKED ISSUE #158: Feature Request: Crafting upgrade in Interfaces
Would be nice to put a crafting upgrade into an interface so it can request an item to be crafted for export

dunkfordyce:
… items. fixes #158

### DIFF
--- a/src/main/java/refinedstorage/tile/TileInterface.java
+++ b/src/main/java/refinedstorage/tile/TileInterface.java
@@ -4,6 +4,7 @@ import io.netty.buffer.ByteBuf;
 import net.minecraft.inventory.Container;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTTagList;
 import net.minecraft.util.EnumFacing;
 import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.items.CapabilityItemHandler;
@@ -18,6 +19,8 @@ import refinedstorage.inventory.BasicItemHandler;
 import refinedstorage.inventory.BasicItemValidator;
 import refinedstorage.item.ItemUpgrade;
 import refinedstorage.tile.config.ICompareConfig;
+
+import java.util.ArrayList;
 
 public class TileInterface extends TileNode implements ICompareConfig {
     private static final String NBT_COMPARE = "Compare";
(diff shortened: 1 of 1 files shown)


raoulvdberge:
Does it also work when you have multiple items set up for exporting?

raoulvdberge:
Also, the crafting task scheduler has to be written to NBT

dunkfordyce:
Ah! Not as easy as I thought! Digging...

raoulvdberge:
You could set up an array of 9 crafting task schedulers, I think.

dunkfordyce:
I am not sure of the best way to write the NBT tags with an array of CraftingTaskSchedulers. I think I should use a NBTTagList but I would have to create a different implementation of CraftingTaskSchedulare.writeToNBT because currently it expects to write to an NBTCompoundTag

raoulvdberge:
Something like this:
```java
NBTTagList l = new  NBTTagList();
for (int i = 0; i < 9; ++i)
{
   if (schedulers[i] == null) continue;
   NBTTagCompound t = new NBTTagCompound();
   t.setInteger("Id", i);
   NBTTagCompound x = new NBTTagCompound();
   schedulers[i].writeToNBT(x);
   t.setTag("Scheduler", x);
   l.appendTag(t);
}
```

dunkfordyce:
Is the code at https://github.com/raoulvdberge/refinedstorage/blob/mc1.10/src/main/java/refinedstorage/apiimpl/autocrafting/CraftingTaskScheduler.java#L40 correct? It sets the scheduled item but should it not also call something like https://github.com/raoulvdberge/refinedstorage/blob/mc1.10/src/main/java/refinedstorage/apiimpl/autocrafting/CraftingTaskScheduler.java#L24 to actually request the crafting?

raoulvdberge:
CraftingTaskScheduler doesn't need to handle calling `addCraftingTask`, the controller calls that already because it saves the crafting state.

raoulvdberge:
I'll close this for now, since it is kinda tricky to implement indeed. Gonna try implementing this myself soon.

raoulvdberge:
(Also, there are some bugs with the current CraftingTaskScheduler, I'll probably have to rework the craft on request system a bit)

dunkfordyce:
Yea I couldn't get it to work! Sorry!

raoulvdberge:
No problem & no need to excuse yourself, guess what, I couldn't get it to work myself :P

PR #179, opened 2016-07-10
Update CommonProxy.java Use ore dictionary for glass blocks instead of vanilla only.

lilrichy:
Use ore dictionary for glass like you do for silicon for better mod compatability.


raoulvdberge:
Doesn't this require `ShapedOreRecipe` for recipes that use oredict?

raoulvdberge:
fixed it myself in master, thanks anyway for the patch!

PR #241, opened 2016-08-02
Adding German translation

ChillUpX:
I took some time and created this german translation. I double checked
ingame, if everything is as it should be :)


raoulvdberge:
Very nice, thank you!

I'll be adding you to CREDITS.md ;)

PR #293, opened 2016-08-25
Added fluid texts to the German language file

0blu:
Take advantage of the preposition "für"


raoulvdberge:
Thank you, however I noticed one slight mistake. Fluid storage is tiered 64K/128K/256K/512K/creative, in the changes you made it tiers like 1K/4K/16K/64K/creative.

PR #316, opened 2016-09-04
Fix ArrayIndexOutOfBoundsException in satisfiedInsertion

TheSeven:
This should fix a crash that occurs if a crafting pattern produces more outputs than it consumes inputs.
Should also fix a possible deadlock with isDone returning false.
This will NOT fix the crash if the bug has already written wrong-length arrays to NBT data, that would need a separate workaround.


raoulvdberge:
Whoops, thanks!

raoulvdberge:
satisfiedInsertion is only present in the dev builds, so it shouldn't matter.

PR #320, opened 2016-09-07
Add method to get the missing items from an ICraftingTask

rubensworks:
I currently use some very nasty reflection to calculate a task's missing items: https://github.com/CyclopsMC/IntegratedDynamics/commit/ba9360e537ab31bc9889c89bdded9cdbc24f5c57
This PR makes it possible to clean this up.

Other future API users might also benefit from this.

Currently, the implementation for `getMissingItems` has been duplicated for `CraftingTaskNormal` and `CraftingTaskProcessing`.
It would be possible to move the 3 boolean arrays to `CraftingTask`, and move `getMissingItems` there as well for a better abstraction, but I haven't done this here, maybe you had a certain reason to do it like this.

### DIFF
--- a/src/main/java/refinedstorage/api/autocrafting/task/ICraftingTask.java
+++ b/src/main/java/refinedstorage/api/autocrafting/task/ICraftingTask.java
@@ -1,11 +1,13 @@
 package refinedstorage.api.autocrafting.task;
 
+import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.world.World;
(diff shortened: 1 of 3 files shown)


raoulvdberge:
This is not generic enough IMO, and since RS wouldn't use that method in the codebase itself I can't justify merging this.

What I'd rather do is move CraftingTask to the API package, and move the satisfied stuff there too.

rubensworks:
I can live with that :-)

raoulvdberge:
Done! (dfb94f05802ccc58c396c52eb912b0bafe976498)

PR #400, opened 2016-09-30
Pattern edit made easier

qMrx3D:
Autocrafting pattern editing made easier
Where Grid and ProcessingPatternEncoder can read an already configured pattern from the output slot.

### DIFF
--- a/src/main/java/refinedstorage/container/ContainerGrid.java
+++ b/src/main/java/refinedstorage/container/ContainerGrid.java
@@ -2,6 +2,7 @@ package refinedstorage.container;
 
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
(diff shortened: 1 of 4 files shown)


raoulvdberge:
I don't understand what this does, mainly because you totally ignored the coding style.

Also, you can already override patterns by just placing them again in the input slot, not merging.

qMrx3D:
Thank you, for looking at the request. I will of course fix the coding style to match.
And for what this change does, it enables the encoders (Grid/Processing) to read existing patterns by placing them in the output slot, making it easy to change that one copper bar that was not the right one for crafting, or changing some item that required some odd nbt value.

And i am sorry for my bad english, english isn't my first language.

raoulvdberge:
I want to keep the behavior simple, which is inputs on the top and outputs on the bottom. No weird overriding of output slots.

Issue #420, opened 2016-10-05
Provide a Voiding Option for Disk Storage


Sorbe:
#### Issue description:
[Feature Request]
Please consider adding a storage option to "void" excess materials / fluids once a storage device is full.
I cannot tell you how many times, I've added void upgrades to barrels and drawers to deal with excess materials -- I can safely say that this is a very useful storage feature.

Darkere:
This is already possible with a detector and an exporter.

choekstr:
There is a 1.8.9 version and I hear there Azanor is actively working on the 1.10.2 version.  Someone said about a month away but I wouldn't take that as definitive.

raoulvdberge:
- [x] Item storage tiles
- [ ] Fluid storage tiles
- [ ] Disk drive

raoulvdberge:
Implemented

Issue #251, opened 2016-08-09
[0.8.15] Storage Drawer Quantities not updating properly on Void Drawers


oitsjustjose:
[Footage of the bug](http://puu.sh/qv15Y/69262412bd.mp4)

Best described by the video really. 

- Refined Storage 0.8.15

- Forge 12.18.1.2046

- Storage Drawers 3.2.1

My storage network doesn't seem to be taking into account the quantities of items if they're going to a void drawer and being voided after a certain quantity

raoulvdberge:
I know the cause of this but won't be able to fix it for the next release.

Luckily, this is no dupe bug. It just displays the wrong count. If it gets too annoying you can just break + replace the controller.

oitsjustjose:
Right, or do what I did in the video :) an extra utilities user might even
be an automated way to fix this.

raoulvdberge:
Love the look of your base btw :p

NarigudoBR:
the user has to be in activate block, right?

oitsjustjose:
I'm actually experiencing a bug with ExU2 Grid Power right now, so I couldn't even test unfortunately :(

PR #430, opened 2016-10-06
Added an option to storage blocks to void items

geldorn:
**Feature description:**
This adds an option to storage blocks to void excess items (if the storage of the block is full it will simply destroy all items which would otherwise be inserted). This should be especially useful for auto generated items where you just want a specific amount of them (cobblestone comes to mind :) ). It is at least a partial implementation of #420 - I just added it to blocks only as i think it is more suitable for them, yet enabling it on other storage systems too should be no problem.

With the correct priority and white list settings you can keep a clean storage system without the fear of overflow.

**Notes:**
For the correct operation I also had to fix the problem that destroyed items are still counted to the cached network state. To counter this problem, I changed the ItemStack returned from the ItemStorageNBT.insertItem implementation to return a negative number if items where voided. In the TileController I also added logic to handle the negative ItemStacks correctly. I think the problem in #251 should be fixable with this too.

The icon added for the feature is only a quick draft as I'm not much of a designer.

<hr/>

**Screenshot of the void mode in action:**
image


raoulvdberge:
Thank you for the very good PR. Like that you fixed that issue with the storage count increasing!

Some remarks:

- Can you update your code so it respects the code style?
- Are the `VOID_EXCESS_OFF` and `VOID_EXCESS_ON` constants really needed?
- Small nit: change the translation to say "Void excess items" with a "yes" and "no" option instead of "on" and "off".

geldorn:
Will do that tomorrow, the constants are not required.
Which coding style issues do you mean? I don't see any available coding style guide for the project.

raoulvdberge:
There is no coding style guide, so that means you need to replicate the existing coding style.

geldorn:
Well i tried to mimic it already, can you give me hints about the problems please?

raoulvdberge:
Your if/else style here:
https://github.com/raoulvdberge/refinedstorage/pull/430/files#diff-2eb04851de70f77dda8d21e3d8bb9be5R533

Method decls on the same line here:
https://github.com/raoulvdberge/refinedstorage/pull/430/files#diff-d3739d17fca91ed0895f87eaff6debdcR309

geldorn:
My IDE was folding them all to one line, this is why I didn't see it, coding style and naming should be fixed with the last commits.

raoulvdberge:
Almost good:

- sidebutton.refinedstorage:void_excess.mode.yes/no can be removed, MC already has gui.yes / gui.no I think.
- There are still some coding style issues: spaces after if/ else: `if (` instead of `if(`

raoulvdberge:
Thank you!

geldorn:
With pleasure :)
I missed one if though, file TileStorage.java line 157

PR #454, opened 2016-10-11
Fix for void upgraded storage drawers using the previous developed void handling.

geldorn:
Also a fix for #251.

### DIFF
--- a/src/main/java/refinedstorage/tile/externalstorage/ItemStorageDrawer.java
+++ b/src/main/java/refinedstorage/tile/externalstorage/ItemStorageDrawer.java
@@ -49,8 +49,7 @@ public class ItemStorageDrawer extends ItemStorageExternal {
             if (!simulate && remainingSpace > 0) {
                 if (drawer.isEmpty()) {
                     drawer.setStoredItemRedir(stack, inserted);
-                }
-                else {
+                } else {
                     drawer.setStoredItemCount(stored + inserted);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/refinedstorage/tile/externalstorage/ItemStorageDrawer.java
```diff
@@ -41,47 +41,31 @@ private boolean isVoidable() {
...
+                    drawer.setStoredItemRedir(stack, inserted);
                 }
-            } else {
-                if (getStored() + size > drawer.getMaxCapacity(stack)) {
-                    int remainingSpace = getCapacity() - getStored();
-
-                    if (remainingSpace <= 0) {
-                        return isVoidable() ? null : ItemHandlerHelper.copyStackWithSize(stack, size);
-                    }
+                else {
```
raoulvdberge: coding style: else should be on same line

geldorn: It should be this is just looking strange in the diff.

raoulvdberge: It's not. Look at the file in your repo: https://github.com/geldorn/refinedstorage/blob/9c38c2998333cb4ab9a33d973ce29aa487bdfb9d/src/main/java/refinedstorage/tile/externalstorage/ItemStorageDrawer.java#L53-L55

geldorn: Was reading line and thinking column ;) Fixed.

raoulvdberge:
Thank you!

PR #462, opened 2016-10-12
Fullness detection when crafting

InusualZ:


src/main/java/refinedstorage/tile/grid/TileGrid.java
```diff
@@ -257,7 +254,10 @@ public void onCrafted(EntityPlayer player) {
             if (i < remainder.length && remainder[i] != null) {
                 if (slot != null && slot.stackSize > 1) {
                     if (!player.inventory.addItemStackToInventory(remainder[i].copy())) {
-                        network.insertItem(remainder[i].copy(), remainder[i].stackSize, false);
+                        ItemStack remain = network.insertItem(remainder[i].copy(), remainder[i].stackSize, false);
```
raoulvdberge: varname should be remainder not remain

src/main/java/refinedstorage/tile/grid/TileGrid.java
```diff
@@ -257,7 +254,10 @@ public void onCrafted(EntityPlayer player) {
...
+                        if (remain != null) {
+                            InventoryHelper.spawnItemStack(player.worldObj, pos.getX(), pos.getY(), pos.getZ(), remain);
```
raoulvdberge: should it spawn at the player pos and not grid pos?

src/main/java/refinedstorage/tile/grid/TileGrid.java
InusualZ: Look up. ItemStack Array called remainder

way2muchnoise: remainderStack?

PR #465, opened 2016-10-14
Destructor can now pickup item from the world

InusualZ:


src/main/java/refinedstorage/tile/TileDestructor.java
```diff
@@ -71,9 +95,29 @@ public int getEnergyUsage() {
...
 
+            BlockPos front = pos.offset(getDirection());
+
+            if (pickupItem && type == IType.ITEMS) {
+                List<Entity> droppedItem = new ArrayList<>();
+
+                Chunk chunk = worldObj.getChunkFromBlockCoords(front);
+                chunk.getEntitiesWithinAABBForEntity(null, new AxisAlignedBB(front), droppedItem, null);
+
+                if (droppedItem.size() > 0) {
```
raoulvdberge: is this check really needed?

src/main/java/refinedstorage/tile/TileDestructor.java
```diff
@@ -71,9 +95,29 @@ public int getEnergyUsage() {
     @Override
     public void updateNode() {
         if (ticks % upgrades.getSpeed(BASE_SPEED, 4) == 0) {
-            if (type == IType.ITEMS) {
-                BlockPos front = pos.offset(getDirection());
 
+            BlockPos front = pos.offset(getDirection());
+
+            if (pickupItem && type == IType.ITEMS) {
+                List<Entity> droppedItem = new ArrayList<>();
```
raoulvdberge: droppedItem**s**

src/main/java/refinedstorage/tile/TileDestructor.java
```diff
@@ -71,9 +95,29 @@ public int getEnergyUsage() {
...
+
+                Chunk chunk = worldObj.getChunkFromBlockCoords(front);
+                chunk.getEntitiesWithinAABBForEntity(null, new AxisAlignedBB(front), droppedItem, null);
+
+                if (droppedItem.size() > 0) {
+                    for (Entity item : droppedItem) {
+                        if (item instanceof EntityItem) {
+                            ItemStack remainder = network.insertItem(((EntityItem) item).getEntityItem().copy(), ((EntityItem) item).getEntityItem().stackSize, false);
+                            if (remainder != null) {
+                                InventoryHelper.spawnItemStack(worldObj, front.getX(), front.getY(), front.getZ(), remainder);
```
raoulvdberge: don't insert and then respawn an item when there is no space

just simulate an  insert first and check if the insert was successful before really inserting and removing the item

src/main/java/refinedstorage/tile/TileDestructor.java
```diff
@@ -71,9 +95,29 @@ public int getEnergyUsage() {
...
+            if (pickupItem && type == IType.ITEMS) {
+                List<Entity> droppedItem = new ArrayList<>();
+
+                Chunk chunk = worldObj.getChunkFromBlockCoords(front);
+                chunk.getEntitiesWithinAABBForEntity(null, new AxisAlignedBB(front), droppedItem, null);
+
+                if (droppedItem.size() > 0) {
+                    for (Entity item : droppedItem) {
+                        if (item instanceof EntityItem) {
+                            ItemStack remainder = network.insertItem(((EntityItem) item).getEntityItem().copy(), ((EntityItem) item).getEntityItem().stackSize, false);
```
raoulvdberge: shouldn't you check `IFilterable.canTake` before running this?

src/main/java/refinedstorage/tile/TileDestructor.java
way2muchnoise: The for each following the statement will not loop when size is Zero. So no it isn't needed

src/main/java/refinedstorage/tile/TileDestructor.java
```diff
@@ -99,20 +99,24 @@ public void updateNode() {
...
-                        if (item instanceof EntityItem) {
-                            ItemStack remainder = network.insertItem(((EntityItem) item).getEntityItem().copy(), ((EntityItem) item).getEntityItem().stackSize, false);
-                            if (remainder != null) {
-                                InventoryHelper.spawnItemStack(worldObj, front.getX(), front.getY(), front.getZ(), remainder);
+                chunk.getEntitiesWithinAABBForEntity(null, new AxisAlignedBB(front), droppedItems, null);
+
+                for (Entity item : droppedItems) {
+                    if (item instanceof EntityItem) {
+                        if (IFilterable.canTake(itemFilters, mode, compare, ((EntityItem) item).getEntityItem())) {
+                            int originalSize = ((EntityItem) item).getEntityItem().stackSize;
```
raoulvdberge: I'm not sure I like this approach, it's not guaranteed that the remainder value is the same item as the drop.

I recommend just blocking the removeEntity if the simulate remainder != null.

raoulvdberge: Also, you seem to be using ((EntityItem) item).getEntityItem() a lot, please extract that in a variable

PR #477, opened 2016-10-16
Render improvements for CraftingMonitorElements and CraftingPreviewElements

way2muchnoise:
Adds fluids to the Preview and gets rid of the weird generic to internal class structure of the rendering


src/main/java/refinedstorage/apiimpl/util/Comparer.java
```diff
@@ -18,7 +18,9 @@ public boolean isEqual(ItemStack left, ItemStack right, int flags) {
         }
 
         if ((flags & COMPARE_OREDICT) == COMPARE_OREDICT) {
-            return isEqualOredict(left, right);
+            if (isEqualOredict(left, right)) {
```
InusualZ: why not `return isEqualOredict(left, right);` ?

InusualZ: nvm, I see why

way2muchnoise: If items aren't registered in the ore dict that will return false even if both stacks are the same Item, Damage and NBT

acidjazz:
lets get a release going! :)

PR #542, opened 2016-10-29
Fixing a common bug on EU Supporters

Speiger:
As i told in the doc. IC2 has tiers in EU. The limit is 0-13 if you go any higher or lower then you start to reset.
Basicly Anything higher is overflowing a Integer also causes longer math etc.
HV i think is the perfect voltage for 1 reason: EU Has energy Packets in Both IC2s. That means You can send in 10 Million EU per tick but these are only HV Packets so nothing will explode. So if anyone tries to do this bla bla bla...
Anyway if you think 3 is to low. Go higher but stay in range of 0-13 Classic would reset you to LV if you do that...


raoulvdberge:
Can you remove all those comments you added? They are unneeded.

Speiger:
Are you fine with HV? but yeah doing it one sec

PR #650, opened 2016-11-19
New API method to allow to connect graphically other tiles to cables

davidog23:
I was writing a open computer driver for Refined Storage and I need to connect graphically the cables to another tile which doesn't implement INetworkNode nor INetworkMaster.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/IRSAPI.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/IRSAPI.java
@@ -89,4 +89,9 @@ public interface IRSAPI {
      * @return a hashcode for the given stack
      */
     int getFluidStackHashCode(FluidStack stack);
+
+    /**
+     * @param classes
+     */
+    void addConnectables(Class<?> classes);
 }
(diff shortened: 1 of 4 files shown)


raoulvdberge:
Thanks for the PR. However I don't think a `Set<Class>` is flexible enough. Can we perhaps change that to a `Set<Predicate<TileEntity>>`? Also, the `connectableElements` field in `BlockCable` is confusing. Can we centralize that to the API?

davidog23:
Something like this?

way2muchnoise:
Wouldn't it be easier to just make some interface like `IConnectable` which has that one method `shouldConnect` or  `canConnect` or maybe not even have a method? (I guess default it to true, so you can have some requirements in your TE)

davidog23:
The problem with that is the fact I must to connect the TE to the cable without modifying the TE signature.
Creating IConnectable would be easier but more restrictive with code you can't modify.

way2muchnoise:
Any reason you can't change the TE signature? Or what is the use case here?

davidog23:
The TE I want to connect is the Adapter from Open Computers and the use I propose is:
`@RSAPIInject
public static IRSAPI RSAPI;`
Inside my init:
`RSAPI.getConnectableConditions().add(tile -> tile instanceof Adapter);`

way2muchnoise:
Being a cable might even better just be a capability. Will have to discus this with @raoulvdberge some time.

raoulvdberge:
Yep, a CableConnector capability would be cool. Are you able to bind such a cap to the adapter, @davidog23?

raoulvdberge:
On the topic of the PR, it's nearly perfect. Can you move the ` add(tile -> tile instanceof INetworkMaster || tile instanceof INetworkNode)` call to `ProxyCommon`. Other API interactions are happening there too.

raoulvdberge:
Thank you!

davidog23:
I was reading about the capability system but It confuses me, but I think I would be able to bind the TE with the CableConnector capability.

Issue #644, opened 2016-11-17
INetworkNode needs to be a capability


raoulvdberge:

Volune:
Hi @raoulvdberge 

I was having a look at this issue (before I try to have a look at enderio cables), and I'd like to suggest a few refactoring on the `INetworkNode`:
- Remove `getNodeWorld()` and add `getDimension()` to `IWirelessTransmitter`
- Remove `canConduct(...)` and use `tile.hasCapability(...)` instead (it has face check)
- Remove `getPosition()` and replace with `getItem()`

And at last one I'm not really sure about:
- Remove `isConnected()`, replace with getNetwork() != null


I can work on a pull request for that if that helps.

raoulvdberge:
How is getItemStack() a valid replacement for getPosition()?

getNodeWorld() is needed

Volune:
`getNodeWorld()` is only used to get the dimension of a IWirelessTransmitter. That's why I'd like to remove it after putting `getDimension()`

`getPosition()` is only used to get the item of a node, for displaying in the controller UI, so I think a `getItemStack()` would be more appropriate - maybe with a better name.

raoulvdberge:
Ah. Good point. You were ahead of me there :)

In that case, feel free to work on the PR but wait til I make the 1.11 branch in that case (which should be later today).

Volune:
My current work is [there](/Volune/refinedstorage/tree/inetworknode-capability)
I'm waiting your new branch to rebase and open a pull request.

The changes were pretty big regarding the refactoring of `isConnected()`, and rework of `rebuild()` to integrate the capability. By the way, is there an actual need to `getConnectableConditions()` ?

raoulvdberge:
At first sight, most changes seem OK (I like the network graph refactor). However, a proper PR will be needed for further review.

For the justification of the connectable conditions see #650

raoulvdberge:
@Volune The 1.11 branch is now public

PR #687, opened 2016-11-28
[1.11] INetworkNode needs to be a capability

Volune:
As discussed in #644, also include various refactoring.

I'd like to make another PR for 1.10, but let's discuss this one first, then then I'll have to see how to avoid breaking changes in 1.10.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
@@ -14,5 +14,4 @@ public interface INetworkNeighborhoodAware {
     interface Operator {
         void apply(World world, BlockPos pos, EnumFacing side);
     }
-
 }
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
@@ -1,0 +1,18 @@
+package com.raoulvdberge.refinedstorage.api.network;
+
+import net.minecraft.util.EnumFacing;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+/**
+ * Allow the network nodes to implement more optimal or non-regular discovery of neighbor nodes
+ */
+public interface INetworkNeighborhoodAware {
+    void walkNeighborhood(Operator operator);
+
+    @FunctionalInterface
+    interface Operator {
+        void apply(World world, BlockPos pos, EnumFacing side);
+    }
+
+}
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
@@ -141,16 +141,18 @@ public class NetworkNodeGraph implements INetworkNodeGraph {
 
     private void removeOtherControler(World world, BlockPos otherControllerPos) {
         if (!controller.getPos().equals(otherControllerPos)) {
-            world.setBlockToAir(otherControllerPos);
-
             IBlockState state = world.getBlockState(otherControllerPos);
+
+            ItemStack itemStackToSpawn = ItemBlockController.createStackWithNBT(new ItemStack(RSBlocks.CONTROLLER, 1, state.getBlock().getMetaFromState(state)));
+
+            world.setBlockToAir(otherControllerPos);
 
             InventoryHelper.spawnItemStack(
                     world,
                     otherControllerPos.getX(),
                     otherControllerPos.getY(),
                     otherControllerPos.getZ(),
-                    ItemBlockController.createStackWithNBT(new ItemStack(RSBlocks.CONTROLLER, 1, state.getBlock().getMetaFromState(state)))
+                    itemStackToSpawn
             );
         }
     }
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
@@ -177,7 +177,7 @@ public class NetworkNodeGraph implements INetworkNodeGraph {
             if (node instanceof INetworkNeighborhoodAware) {
                 ((INetworkNeighborhoodAware) node).walkNeighborhood(operator);
             } else {
-                for (EnumFacing checkSide : EnumFacing.values()) {
+                for (EnumFacing checkSide : EnumFacing.VALUES) {
                     if (checkSide != side) { // Avoid going backward
                         INetworkNode nodeOnSide = NETWORK_NODE_CAPABILITY.cast(tile.getCapability(NETWORK_NODE_CAPABILITY, checkSide));
                         if (nodeOnSide == node) {
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
@@ -7,6 +7,7 @@ import com.raoulvdberge.refinedstorage.api.network.INetworkNodeGraph;
 import com.raoulvdberge.refinedstorage.api.storage.fluid.IFluidStorageProvider;
 import com.raoulvdberge.refinedstorage.api.storage.item.IItemStorageProvider;
 import com.raoulvdberge.refinedstorage.item.ItemBlockController;
+import com.raoulvdberge.refinedstorage.proxy.CapabilityNetworkNode;
 import com.raoulvdberge.refinedstorage.tile.TileController;
 import com.raoulvdberge.refinedstorage.tile.TileNetworkTransmitter;
 import net.minecraft.block.state.IBlockState;
@@ -118,7 +119,7 @@ public class NetworkNodeGraph implements INetworkNodeGraph {
             }
 
             for (EnumFacing facing : EnumFacing.VALUES) {
-                if (node.canConduct(facing)) {
+                if (tile instanceof TileController || tile.hasCapability(CapabilityNetworkNode.NETWORK_NODE_CAPABILITY, facing)) {
                     BlockPos pos = currentPos.offset(facing);
 
                     if (checked.add(pos)) {
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
@@ -57,10 +57,10 @@ public class NetworkNodeGraph implements INetworkNodeGraph {
             }
         };
 
-        BlockPos controlerPos = controller.getPos();
+        BlockPos controllerPos = controller.getPos();
         World controlerWorld = controller.getWorld();
         for (EnumFacing facing : EnumFacing.VALUES) {
-            BlockPos pos = controlerPos.offset(facing);
+            BlockPos pos = controllerPos.offset(facing);
             operator.apply(controlerWorld, pos, facing.getOpposite());
         }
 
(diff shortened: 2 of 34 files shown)


src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
```diff
@@ -30,11 +31,7 @@ public NetworkNodeGraph(TileController controller) {
     }
 
     @Override
-    public void rebuild(BlockPos start, boolean notify) {
-        if (start == null) {
-            start = controller.getPosition();
-        }
-
+    public void rebuild() {
```
raoulvdberge: I think you forgot about the interdimensional network system in your refactor, as I don't think it works anymore.

src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
```diff
@@ -0,0 +1,18 @@
...
+import net.minecraft.world.World;
+
+/**
+ * Allow the network nodes to implement more optimal or non-regular discovery of neighbor nodes
+ */
+public interface INetworkNeighborhoodAware {
+    void walkNeighborhood(Operator operator);
+
+    @FunctionalInterface
+    interface Operator {
```
raoulvdberge: Can we have a better name for this?

src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
```diff
@@ -0,0 +1,18 @@
...
+        void apply(World world, BlockPos pos, EnumFacing side);
+    }
+
```
raoulvdberge: Unneeded newline

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/item/NetworkItemHandler.java
```diff
@@ -36,7 +36,7 @@ public void onOpen(EntityPlayer player, World controllerWorld, EnumHand hand) {
         boolean inRange = false;
 
         for (INetworkNode node : network.getNodeGraph().all()) {
-            if (node instanceof IWirelessTransmitter && node.getNodeWorld().provider.getDimension() == player.dimension) {
+            if (node instanceof IWirelessTransmitter && ((IWirelessTransmitter)node).getDimension() == player.dimension) {
```
raoulvdberge: Coding style: `((INetworkTransmitter) node)`

src/main/java/com/raoulvdberge/refinedstorage/proxy/CapabilityNetworkNode.java
```diff
@@ -0,0 +1,62 @@
...
+import net.minecraft.util.EnumFacing;
+import net.minecraftforge.common.capabilities.Capability;
+import net.minecraftforge.common.capabilities.CapabilityInject;
+import net.minecraftforge.common.capabilities.CapabilityManager;
+
+import javax.annotation.Nullable;
+
+public class CapabilityNetworkNode {
+    @CapabilityInject(INetworkNode.class)
+    public static Capability<INetworkNode> NETWORK_NODE_CAPABILITY = null;
```
raoulvdberge: Shouldn't this capability be available in the API package?

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
raoulvdberge: Nevermind, it still does I just see a bit further in your diff. :)

src/main/java/com/raoulvdberge/refinedstorage/api/network/INetworkNeighborhoodAware.java
raoulvdberge: Unneeded newline

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
```diff
@@ -200,4 +138,49 @@ public void disconnectAll() {
...
+            this.side = side;
+            this.tile = tile;
+        }
+
+        @Override
+        public void walkNeighborhood(Operator operator) {
+            if (node instanceof INetworkNeighborhoodAware) {
+                ((INetworkNeighborhoodAware) node).walkNeighborhood(operator);
+            } else {
+
```
raoulvdberge: Unneeded newline

src/main/java/com/raoulvdberge/refinedstorage/gui/grid/GuiGrid.java
```diff
@@ -201,7 +201,11 @@ private int getRows() {
     }
 
     private boolean isOverSlotWithItem() {
-        return grid.isConnected() && isOverSlot() && slotNumber < STACKS.size();
+        return isGridActive() && isOverSlot() && slotNumber < STACKS.size();
+    }
+
+    private boolean isGridActive() {
```
raoulvdberge: Is this method really needed? Can't we just call `grid.isActive()` immediatally?

raoulvdberge:
Very good PR. Thank you. If the points I brought up are resolved, I will merge.

src/main/java/com/raoulvdberge/refinedstorage/proxy/CapabilityNetworkNode.java
Volune: Having INetworkNode in the API should be enough, other mods will use `@CapabilityInject(INetworkNode.class)`. I'm trying a bit in enderio to ensure it will integrate as expected.

src/main/java/com/raoulvdberge/refinedstorage/gui/grid/GuiGrid.java
Volune: It made the refactoring easier, I'll inline it.

Volune:
PR updated.

I still have one concern: when trying to integrate with EnderIO, the conduits won't connect with the Controller. (RS cables and other RS tiles are fine)
I think we should add the capability to the controller and make it extends INetworkNode.

raoulvdberge:
Yes, that might be a good idea. You want to do it for this PR or for another PR?

raoulvdberge:
Then I have another question regarding the capabilities: is it possible to bind caps to TE's you don't have access too? If that is the case, we can remove the connectables system and move it to a `IConnector` capability or something.

Volune:
It's possible to bind capabilities to other TileEntity, for example here is a piece of code I'm using to build EnderIO RS conduits:
```java
public class RSConduitCapabilityProvider implements ICapabilityProvider {
    private static final ResourceLocation KEY = new ResourceLocation(EnderIO.DOMAIN, "EioCapRSNetworkNode");

    @CapabilityInject(INetworkNode.class)
    public static Capability<INetworkNode> NETWORK_NODE_CAPABILITY = null;

    private final TileConduitBundle tile;

    public RSConduitCapabilityProvider(TileConduitBundle tile) {
        this.tile = tile;
    }

    @SubscribeEvent
    public static void attachCapabilities(AttachCapabilitiesEvent<TileEntity> event) {
        Object object = event.getObject();
        if (object instanceof TileConduitBundle) {
            event.addCapability(KEY, new RSConduitCapabilityProvider((TileConduitBundle) object));
        }
    }

    @Override
    public boolean hasCapability(Capability<?> capability, @Nullable EnumFacing facing) {
        if (capability == NETWORK_NODE_CAPABILITY) {
            IRSConduit conduit = tile.getConduit(IRSConduit.class);
            return conduit != null && !conduit.isRemoved() &&
                    conduit.getConnectionMode(facing) == ConnectionMode.IN_OUT;
        }
        return false;
    }

    @Override
    public <T> T getCapability(Capability<T> capability, @Nullable EnumFacing facing) {
        if (capability == NETWORK_NODE_CAPABILITY) {
            IRSConduit conduit = tile.getConduit(IRSConduit.class);
            if (conduit != null && !conduit.isRemoved() &&
                    conduit.getConnectionMode(facing) == ConnectionMode.IN_OUT) {
                return NETWORK_NODE_CAPABILITY.<T>cast(conduit.getRSNetworkNode());
            }
        }
        return null;
    }
}
```

I'm wondering in which case we would want a `IConnector` not to be a `INetworkNode`. Having to check for only one capability will make integration easier.

I'll make the Controller a `INetworkNode` in this PR.

raoulvdberge:
I see, cool.

There is 1 remaining problem I see.

When a `INetworkNode` is placed, it needs to this to refresh the graph of the network of its neighbors:

```
    @Override
    public void onBlockPlacedBy(World world, BlockPos pos, IBlockState state, EntityLivingBase player, ItemStack stack) {
        super.onBlockPlacedBy(world, pos, state, player, stack);

        if (!world.isRemote) {
            for (EnumFacing facing : EnumFacing.VALUES) {
                TileEntity tile = world.getTileEntity(pos.offset(facing));

                if (tile instanceof TileNode && ((TileNode) tile).isConnected()) {
                    ((TileNode) tile).getNetwork().getNodeGraph().rebuild();

                    break;
                }
            }
        }
    }

    @Override
    public void breakBlock(World world, BlockPos pos, IBlockState state) {
        INetworkMaster network = null;

        if (!world.isRemote) {
            TileEntity tile = world.getTileEntity(pos);

            if (tile instanceof TileNode) {
                network = ((TileNode) tile).getNetwork();
            }
        }

        super.breakBlock(world, pos, state);

        if (network != null) {
            network.getNodeGraph().rebuild();
        }
    }
```
This is currently hardcoded in `BlockNode`, we might want to change that with block break / place events.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/NetworkNodeGraph.java
```diff
@@ -43,112 +40,53 @@ public void rebuild(BlockPos start, boolean notify) {
...
-            for (EnumFacing facing : EnumFacing.VALUES) {
-                if (node.canConduct(facing)) {
-                    BlockPos pos = currentPos.offset(facing);
-
-                    if (checked.add(pos)) {
-                        toCheck.add(pos);
-                    }
-                }
-            }
+        BlockPos controlerPos = controller.getPos();
```
raoulvdberge: Typo: contro**l**erPos

raoulvdberge:
I added some minor additional comments.

src/main/java/com/raoulvdberge/refinedstorage/tile/TileController.java
```diff
@@ -919,6 +921,34 @@ public int getEnergyUsage() {
         return usage;
     }
 
+    @Nullable
+    @Override
+    public ItemStack getItemStack() {
```
InusualZ: I don't think this function would ever return null

src/main/java/com/raoulvdberge/refinedstorage/tile/TileController.java
```diff
@@ -105,18 +108,17 @@ public Integer getValue(TileController tile) {
 
             for (INetworkNode node : tile.nodeGraph.all()) {
                 if (node.canUpdate()) {
-                    IBlockState state = tile.getWorld().getBlockState(node.getPosition());
+                    ItemStack itemStack = node.getItemStack();
+                    if (itemStack == null) {
```
InusualZ: ItemStack could be null?

Volune: Yes, could be null. `getItemStack()` is `@Nullable` in the interface. I think it would be more convenient for integration in other mods, and that was also a requirement in my 1.10 branch, where most implementations may return null. Only the default capability may return null in this pull request.

Volune:
Regarding the Place/Break events, I don't think we can use them. They are cancelable (and may be canceled by another mod), but we want the effect only when the block is actually placed/broken. I don't know if we can do that with the events.
Also I'm concerned about the performances, if we have to check the capabilities for all blocks.

src/main/java/com/raoulvdberge/refinedstorage/tile/TileController.java
InusualZ: That's why I'm saying it. I think that you should return Air ItemStack with consistency with 1.11

raoulvdberge: Yes, `ItemStack.EMPTY`

src/main/java/com/raoulvdberge/refinedstorage/tile/TileController.java
```diff
@@ -109,7 +110,7 @@ public Integer getValue(TileController tile) {
             for (INetworkNode node : tile.nodeGraph.all()) {
                 if (node.canUpdate()) {
                     ItemStack itemStack = node.getItemStack();
-                    if (itemStack == null) {
+                    if (itemStack.getItem() == Items.AIR) {
```
raoulvdberge: No, use `itemStack.isEmpty()` instead

raoulvdberge:
Great, I'll merge. Thanks for this! Network scanning is much cleaner now. And mods can use the cap system super easily.

Feel free to join #refinedstorage on esper.net as well, @Volune

PR #703, opened 2016-12-03
[1.10] INetworkNode needs to be a capability

Volune:
A new pull request on 1.10, based on the one from 1.11.

In this one I tried to avoid any breaking API change (only the `getConnectableConditions` should have changed).

And some teasing, EnderIO integration (which is why I ask this PR): 
2016-12-03_12 07 18_min


raoulvdberge:
Sorry, 1.10 is in bugfix only mode now. We won't accept any refactorings.

PR #731, opened 2016-12-10
Mc1.10 feature request

LertuZUES:
can you please add support ender io [energy,item,fluid]
and also support for mekansim and extra utilities 2


raoulvdberge:
no idea what you're talking about. also, you made a PR instead of a regular issue.

PR #777, opened 2016-12-22
Reader and Writer Aesthetics Changed

Boomflex:
I felt the "IN" and "OUT" looked really unnactracive so I have made it a
uniformed size with I and O.


raoulvdberge:
lol those aren't even the final textures so it doesn't matter

PR #903, opened 2017-01-23
Adds small messages

Ristellise:
This is to tell users that for all freezes that they should post a relavent log for investigation we can't help them if they can't help us. also tell e users to replace the quare brackets with relavent jnfo.


Joshwoo70:
if you don't want it feel free to remove close it.

raoulvdberge:
Current text is clear enough I think.

Issue #969, opened 2017-02-08
[QoL] Processing Pattern Encoder recipe only accepts vanilla workbench, not ore dict compatible ones


kosakriszi:
#### Issue description:
Using ore dictionary compatible workbenches (i.e. Tinkers Construct Crafting Station) in the recipe doesn't work.

#### What happens:
The recipe is not seen as valid.

#### What you expected to happen:
To be able to use all compatible workbenches as ingredients.

#### Version:
Most likely all.

raoulvdberge:
Define "doesn't work".

kosakriszi:
Here's a short video detailing the issue: https://gfycat.com/UnsungRecentAntlion

kosakriszi:
Here is a short video detailing the issue: https://gfycat.com/LamePastelCornsnake

VT-14:
I think the recipe is using only the vanilla Crafting Table, not the ore dictionary equivalents.  Refined Storage is looking for "Blocks.CRAFTING_TABLE"

I also found Blocks.CHEST, which you could use the ore dictionary tag chestWood (I think) to add in compatibility with Quark's variants (among others).

According to that GIF, vanilla has the tags
-workbench
-craftingTableWood

While TiCon has only
-workbench

VT-14:
I made a fork and changed the Crafting Grid, Processing Pattern Encoder, Crafting Upgrade, Disk Drive , and (in a moment, just now saw it) the Security Manager to use the ore dictionary tags "workbench" and "chestWood" as appropriate.  Unfortunately, I won't be able to test this until tonight and have no idea if I did it correctly or not (total programming experience: Introduction to C++ college class and simple Lua programs in Computercraft/OpenComputers).

I did see that the External Storage already uses the "chest" ore dictionary tag.  Is that preferred over chestWood?

raoulvdberge:
yes

VT-14:
I failed to get the Solderer recipes working, but at least I got the vanilla crafting recipes to use the ore dictionary.  The above pull requests should 'fix' the Processing Pattern Encoder and Security Manager.

PR #976, opened 2017-02-10
OreDictionary Support for the PPE and Security Manager.

VT-14:
Added the "workbench" tag for the Crafting Table in the Processing Pattern Encoder, and the "chest" tag for the Security Manager. This allows them to use crafting tables and chests from other mods, including the Crafting Station from Tinkers' Construct and the Chests from Quark. #969

It appears that the Solderer recipes are looking for ItemStacks and do not like the strings for ore dictionary support. I'm not sure how to add the ore dictionary versions to the Crafting Grid, Crafting Upgrade, or Disk Drive.


raoulvdberge:
thanks!

PR #1080, opened 2017-03-20
Allow INodeNetwork instances to return an ItemStack of their choice for display in Controller GUI

bmwalter68:
Should not break the existing API as the default function implementation is the behavior of the current implementation.


raoulvdberge:
Thanks

PR #1089, opened 2017-03-24
Fix connections conductivity

bmwalter68:
Check that facing blocks can conduct in each others direction when deciding on visually showing connection. (I'm looking at you BlockCable)

Replace equivalent code check in BlockCable that replicated canConduct in TileMultipartNode.

Add a hasConnectableConditions method to the public API to perform the getConnectableConditions stream test for mods that can't/don't do lambda functions. (java 1.7 and lower)  Consume that method in our own tests for block-to-block connectivity.

The RS network visual connectivity should be closer to the functional connectivity.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/IRSAPI.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/IRSAPI.java
@@ -106,4 +106,9 @@ public interface IRSAPI {
      * @return a set with the predicates to check if a block is connectable
      */
     Set<Predicate<TileEntity>> getConnectableConditions();
+
+    /**
+     * @return whether the provided TileEntity matches any of the current connectable conditions
+     */
+    boolean hasConnectableConditions(TileEntity te);
 }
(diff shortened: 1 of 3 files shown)


raoulvdberge:
Why aren't you calling `TileMultipartNode.hasBlockingMicroblock` anymore?

bmwalter68:
That code is identical to canConduct(direction) in TileMultipartNode...  Basically, you were returning canConduct in that case, and since it was also appropriate to call canConduct for all INetworkNode instances that call was condensed into a single canConduct.

raoulvdberge:
Did you test this change with multiparts?

bmwalter68:
I did compare all of the test cases I could think of pre-and-post-change.   Everything I tried worked the same in both cases.

Just to be clear, TileMultipartNode.hasBlockingMicroblock is still getting called.  The outer condition covers all INetworkNodes, which includes TileMultipartNode.  We do the TileMultipartNode specific test for placementType and if that passes we return canConduct.  The canConduct method for TileMultipartNode is:

return !hasBlockingMicroblock(getWorld(), pos, direction) && !hasBlockingMicroblock(getWorld(), pos.offset(direction), direction.getOpposite());

Which is the same as the old fallthrough.  It's still getting called, just from canConduct.

PR #1116, opened 2017-04-01
Fix missed node graph rebuild when placing node blocks on the far side of a conductive addon block implementing INetworkNode

bmwalter68:
The current code in BlockCable.attemptConnect() and BlockNode.onBlockPlacedBy() test the TE's of surrounding blocks to see if they are network nodes and a node graph rebuild should take place on placement and after neighbors change.  The test is against TileNode which restricts the rebuild to adjacent RS blocks only, not addon blocks implementing INetworkNode on their TE.  Because of this, when BlockCable or BlockNode descendents are placed on the far side of a conductive addon block implementing INetworkNode, the node graph is not refreshed as it should be.


raoulvdberge:
Thanks

PR #1118, opened 2017-04-01
Add hasConnectableConditions() to API

bmwalter68:
Add API convenience function to test connectable conditions for mods consuming our API and still targeting java 1.7 for whatever reason (e.g. no lambda functions)


raoulvdberge:
Not needed, you could easily write your own utility function to handle this. Convenience functions don't go in API.

PR #1192, opened 2017-04-29
Fix CraftingTweaks Integration

BlayTheNinth:
Updated to latest Crafting Tweaks API. Buttons will now always appear next to the grid, no matter at what slot index it starts.

Also set `keyHandled = true` in `GuiGrid` to prevent the `KeyboardInputEvent.Post` event from firing, which will fix Crafting Tweaks hotkeys triggering when they shouldn't (Crafting Tweaks v7.1.7+).

What's not fixed is that the Crafting Tweaks buttons disappear when changing the grid size (they re-appear on the next `setWorldAndResolution` call, so on resizing the window or reopening the gui). I don't consider it a huge issue, but will fix that on my side later.


raoulvdberge:
Thanks

PR #1208, opened 2017-05-09
Added more item and fluid methods for OC to access

davenonymous:
This PR actually adds three methods:
- extractItem(stack, size, side)

as discussed.

I've additionally implemented these two to align the fluid functionality with that of items.
- extractFluid(stack, size, side)
- getFluid(stack)

Sadly OC does not have a checkFluidStack method to grab a fluidstack directly from the arguments, so we have to deal with that ourselves - which is only the name, so not too much complexity there. A more elaborate version of this should probably be integrated into OpenComputers directly - but it's scala and that would take me a bit to get into.

The getItem() method is now also able to compare against the oreDictionary.


raoulvdberge:
Thank you!

PR #1282, opened 2017-06-04
fix auto-crafting ore dictionary

Matrix89:


raoulvdberge:
Can you tell me what you did and what issues that this fixes?

Matrix89:
The issue is that when you have a pattern for energy upgrade from mekanism(it requires gold dust) the system says that you don't have gold grit(gold dust equivalent from immersive enginnering) even if you have gold dust.
I added a check if the pattern is OreDicted.

way2muchnoise:
[This is where ore dictionary things get handled](https://github.com/raoulvdberge/refinedstorage/blob/mc1.11/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/task/CraftingTask.java#L122-L145). The issue you have is that mechanism has it's own [OreShapedRecipe](https://github.com/aidancbrady/Mekanism/blob/master/src/main/java/mekanism/common/recipe/ShapedMekanismRecipe.java) and I have to [hack around it](https://github.com/raoulvdberge/refinedstorage/blob/mc1.11/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java#L62). Which might cause issues, what you are doing will cause even more. As in your solution you'll be able to substitute anything as long as it has a shared oredict entry, eg. use any dye when lapis is needed.

It not working has two options, the recipe isn't oredicted correctly or the ingredients you are using aren't oredicted correctly.

PR #1289, opened 2017-06-07
Fix calculating crafting resources for more than 9 unique items

ExpensiveKoala:


way2muchnoise (commented):
I don't really see what is wrong with the current code, a crafting grid can't really be bigger as a 3x3. This code is only used for internal CraftingSteps that involve a vanilla crafting recipe. So it should never be more than 9 stacks.

raoulvdberge:
He's making an addon for the Avaritia crafting table, for 9x9 recipes.

ExpensiveKoala:
The rest of the code in that method is using grid.size() either way. This change wouldn't break anything that was working before but it would allow each craft task to have as many ingredients as the pattern has.

way2muchnoise (changes requested):
Alright, one change tho. Use `Math.max(9, grid.size())` just so it doesn't change any existing behaviour.

way2muchnoise:
@ExpensiveKoala I'll will break [this](https://github.com/raoulvdberge/refinedstorage/blob/mc1.11/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java#L187) and [this](https://github.com/raoulvdberge/refinedstorage/blob/mc1.11/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java#L218) for starters and I have only look for direct uses. It just keeps the old behaviour and will support your new one.

ExpensiveKoala:
It won't break those two parts because in order to have a 9x9 or even any recipe greater than 3x3 it's required to use a custom ICraftingPattern. Unless in the background during auto crafting it tries to wrap the pattern created by the IPatternProvider to CraftingPattern it wouldn't be a problem. When I can, I'll implement the Math.max(9, grid.size()) but  wouldn't matter because the for loop that's right after it would still loop to grid.size() either way. The reason I needed this change is because it tries to loop i to 81, but the ItemStack array was coded to 9. (Index out of bounds)

If you want the Math.max it'll need the math.max everywhere in that method that uses grid.size()

way2muchnoise:
It won't need a `Math.max` everywhere, it just expects an array of nine elements everywhere down the line. Which is why it should never be less, thus needing the `Math.max`. It is to keep the minimum size like it is, it'll only be bigger in your addon anyways.

You are requesting a change to an internal method which you are going to use in an addon. That isn't a common convention, but whatever. This is why I just want to keep the old behaviour intact.

ExpensiveKoala:
I'm not using this method for my add-on at all. It's just this method causes an index out of bounds error with a pattern with more than 9 items. 

But okay. I'll do the Math.max.

raoulvdberge:
@way2muchnoise 

@ExpensiveKoala is using their own pattern implementation that provides a regular RS crafting task, so, the code in `CraftingPattern` wouldn't be affected.

way2muchnoise (approved):

way2muchnoise:
@raoulvdberge I can see that, but the issue is that if the gridsize is 2x2 which is possible in vanilla or even 1x1 the lenght of the array will become 1 or 4. While in the normal crafting pattern 9 outputs are expected. This is why the max is needed.

raoulvdberge:
OK

PR #1367, opened 2017-07-15
[Mc1.12] Add Forestry Bee support to filters

TeknoKit:
The only problem with automating forestry bees is that the queens have an NBT tag that increments their generations, making the result technically a different item. This extra comparison checks to see if the item coming in has this GEN tag, and if so, ignores the GEN tag before comparing everything else.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/util/Comparer.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/util/Comparer.java
@@ -99,10 +99,8 @@ public class Comparer implements IComparer {
                 return true;
             } else if (!left.hasTagCompound() && right.hasTagCompound() && right.getTagCompound().hasNoTags()) {
                 return true;
-            } 	
-	     /* Forestry Bee Support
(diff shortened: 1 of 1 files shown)


way2muchnoise:
Dunno if this is really the best place for this piece of code

TeknoKit:
Another implementation idea was to add a "forestry" domain to the Comparer.stripTags method and remove the "GEN" tag from the item before comparing, but it doesn't seem that exporters are ever set to strip NBT tags in any situation, and I am not sure that would cause other problems.
Editing every node for stripTags also seems bad just for this one case, as a player who would want an exporter to ignore the tag would most likely also want disk filters to ignore the tag.

It's only a suggestion. I see no reasonable situation in which any Forestry player automating bees would ever care about sorting them based on the GEN tag.

raoulvdberge:
I'm not sure about this one either, it's up to you to decide @way2muchnoise.

Either way, if this gets merged, I want a comment that says it's for Forestry bees in the code.

raoulvdberge:
The comment has to appear in the if block of the removeTag, as the coding style.

raoulvdberge:
I don't like this approach, it's not the right place to add that check. Closing.

PR #1370, opened 2017-07-18
Forestry Support

TeknoKit:
At the start of any isEqual comparison, if Forestry is loaded, it determines if either item is a Forestry Bee with a GEN tag and if so flags the comparison with COMPARE_STRIP_NBT. A forestry case with the GEN tag has been added to the stripTags method.
Is this better?


src/main/java/com/raoulvdberge/refinedstorage/integration/forestry/IntegrationForestry.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Determines if any ItemStack in a list contains a Forestry Queen/Princess Bee
+     * and flags it for NBT strip if it contains a "GEN" NBT tag.
+     * 
+     * @param flags	the flags to compare on, see {@link IComparer}
+     * @param items	list of stacks being considered
+     * @return		previous compare flags with COMPARE_STRIP_NBT if any itemstack in list is a 
+     * 				Forestry Queen/Princess Bee with a GEN NBT tag, otherwise only previous compare flags
+     */
+	public static int isItem(int flags, ItemStack... items) {
```
way2muchnoise: This method is kind ambiguous as `isItem` should return a `boolean` indicating whether it is a forestry item or not. Would rather see this method replace with something like a `isBee` method that takes a single `ItemStack`.

raoulvdberge:
No, it's not. The comparer needs to guarantee correct compares, we can't just strip nbt when not asked to strip.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/util/Comparer.java
```diff
@@ -20,6 +21,11 @@ public boolean isEqual(@Nullable ItemStack left, @Nullable ItemStack right, int
         if (validity == EnumActionResult.FAIL || validity == EnumActionResult.SUCCESS) {
             return validity == EnumActionResult.SUCCESS;
         }
+        
+        if(IntegrationForestry.isLoaded()) {
+        	ItemStack[] items = {left, right};
+        	flags = IntegrationForestry.isItem(flags, items);
```
way2muchnoise: Changing the flag here seems kinda off. I would propose adding a method in `IntegrationForestry` along the lines of `sanitise`, where it removes then tags that need to be removed without changing the comparison flags. The this if would be looking more like this:

```java
if (IntegrationForestry.isLoaded()) {
    if (IntegrationForestry.isBee(left) {
        left = IntegrationForestry.sanitise(left.copy());
    }
    if (IntegrationForestry.isBee(right)) {
        right = IntegrationForestry.sanitise(right.copy());
    }
}
```

In this approach `sanitise` could be added to or changed without changing this code.

way2muchnoise: This `isBee` method could be changed to something that would also support trees and butterflies. And name it appropriately, eg. `isBreedable`.

src/main/java/com/raoulvdberge/refinedstorage/integration/forestry/IntegrationForestry.java
way2muchnoise: As mentioned in the other comment it would be nice if this also had support for trees and butterflies. Might as well go all with this support.

way2muchnoise (changes requested):

TeknoKit:
I was under the impression that modifying NBT tags directly outside of comparison was not wanted and thus tried to integrate it to the one method that did such operations but wanted to avoid that for your stated reason the first time. Either way I do understand why it was better to be moved.
I am also sure that moving above the IComparer is a bad idea; trying to ensure all the different items that can extract and insert all accommodate for these items is bound to just get messy and mistakes made.

Yes, I could fairly quickly add in trees and butterflies once I understand the NBTCompounds.
I'll get this right eventually, but I did need some gauging to see the proper mindset here. I'll work on it.

TeknoKit:
I think setting appropriate methods with a sanitize flag is better than altering the default behavior and then trying to blacklist some methods. I'll change that behavior back and put the flags in properly.

way2muchnoise:
This is starting to look at lot better than before, keep up the good work!

TeknoKit:
Should I just re-base the fork at this point?

TeknoKit:
I'll re-fork and start a new pull once this is fixed.

Issue #1233, opened 2017-05-20
Stacks of Air


cjm721:
#### Issue description:
Multiple ItemStacks of Air in Storage
#### What happens:
Do not know what causes it but have been generating them semi often in my base.
#### What you expected to happen:
None to be generated
#### Steps to reproduce:
Do not know at this time

#### Version (make sure you are on the latest version before reporting):
- Minecraft: 1.11.2
- Forge:  13.20.02299
- Refined Storage: 1.4.6

Does this issue occur on a server?  Yes

raoulvdberge:
Which storages are you using? If you are using external storage(s) too, tell me to what they're connected.

cjm721:
1x 4k Storage Block
2x 64k Storage Disk

No external storage

LittleODie:
Also getting this, and a crash when trying to remove the "air" from the system.

The game crashed whilst updating screen events
Error: java.lang.IllegalArgumentException: ItemStack value must not be empty. 0xtile.air@0 minecraft:air

raoulvdberge:
@LittleODie what storages are you using?

LittleODie:
My setup has 2 64k drives. I have a basic setup with one importer connected to a "quantum quarry" from extrautils2. Since the quarry mines so quickly the Importer has 1 stack upgrade and 1 speed upgrade. I thought that the setup might be inputting things too fast causing the issues so I took the speed upgrades out and problem still persisted.

By the way I am getting air & the other issue reported of items not showing up in real-time in the grid. However the real-time issue is intermittent and ceased to be an issue after a few minutes.

knoxz:
also got some air. Disappeared after sleeping  through the night though...
Only using a diskdrive with 4 8k drives.

Matrix89:
Same problem, I have some external storage drawers connected, a few importers(mekanism machines, Environmental Tech miner) and one disk drive full of 64k disks.
- refinedstorage 1.4.16
- Minecraft 1.11.2
- Forge 13.20.0.2310

PessimisticBastard:
Same problem as well, though I am using a few external storage buses hooked up to storage drawer controllers.

thejams13:
I am getting the same thing. Air :) think that maybe Actually Additions  farmer harvested air from pams harvest craft mustard seed.

cjm721:
For me I didn't have AA auto importing into my system.

raoulvdberge:
[...]

PR #1419, opened 2017-08-11
Fixed stacks of air #1233

marcin212:


raoulvdberge:
We don't accept pull requests for mc1.11.

Issue #1511, opened 2017-10-31
[1.12.2] Casue lag on server!


yuanjie000:
#### Issue description:
once using rs to autocraft something, the sever become lag.
used **hundreds of** _crafting pattern_ 

![tps](https://user-images.githubusercontent.com/20902900/32207451-e17e4c98-be36-11e7-8d77-f846f9e08e33.png)

#### Version (make sure you are on the latest version before reporting):
- Minecraft:1.12.2
- Forge:14.23.0.2517
- Refined Storage:refinedstorage-1.5.21

Does this issue occur on a server? [yes/no]
yes
#### If a (crash)log is relevant for this issue, link it here:
here is an nps file of the server
[lowtps.zip](https://github.com/raoulvdberge/refinedstorage/files/1429604/lowtps.zip)

3TUSK:
![image](https://user-images.githubusercontent.com/9113033/32207703-f90ec184-bdba-11e7-90b3-8bbc92a66866.png)

Sorry for jumping in. I was asked for giving a rough analysis on the profiling report. As you can see, I stunk on these abnormal behaviors:
![image](https://user-images.githubusercontent.com/9113033/32207737-260442f4-bdbb-11e7-9016-3b46046b23c1.png)
![image](https://user-images.githubusercontent.com/9113033/32207742-385be7cc-bdbb-11e7-98ff-e74746a25238.png)
![image](https://user-images.githubusercontent.com/9113033/32207750-43ea29b4-bdbb-11e7-8f10-168bdbcd7824.png)

[`isOreDict` should not be that time-consuming.](https://github.com/raoulvdberge/refinedstorage/blob/mc1.12/src/main/java/com/raoulvdberge/refinedstorage/item/ItemPattern.java#L143) Moreover, searching all previous issues in the issue tracker does not reveal any similar issue(s) in the past. The most-matching issue is #838 which has been duplicated multiple times, but according to the profiling result submitted, it is unlikely to connect #838 to this issue. 
Therefore, this issue shall be considered as a separate issue, as it may point out a new underlying design flaw.

raoulvdberge:
A network is turning off and on, causing the crafting pattern chain to rebuild constantly. Make sure enough power is supplied to the controller. This is not caused by autocrafting.

PR #1512, opened 2017-10-31
Fix issue #1511 . Use hashmap to optimize the performance in building CraftingPatternChainList

xinyuan-liu:
Fix issue #1511 .
The original implementation has a complexity of O(n^2) to build a CraftingPatternChainList when a network starts. For each crafting pattern, it need to be compared with all the patterns already in the List, which may cost a very long time when a network has a large amount of patterns. Especially in the server it may cause the players to be disconnected due to the timeout.  With a hash map the complexity can be reduced to O(n), which means at most n times of comparison are needed.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
@@ -14,7 +14,10 @@ public interface ICraftingPattern {
      * @return the {@link ICraftingPatternContainer} where the pattern is in
      */
     ICraftingPatternContainer getContainer();
-
+    
+    public int hashCode();
+    
+    public boolean equals (Object obj);
     /**
      * @return the crafting pattern stack
      */
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
@@ -15,9 +15,6 @@ public interface ICraftingPattern {
      */
     ICraftingPatternContainer getContainer();
     
-    public int hashCode();
-    
-    public boolean equals (Object obj);
     /**
      * @return the crafting pattern stack
      */
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
@@ -14,7 +14,7 @@ public interface ICraftingPattern {
      * @return the {@link ICraftingPatternContainer} where the pattern is in
      */
     ICraftingPatternContainer getContainer();
-    
+
     /**
      * @return the crafting pattern stack
      */
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
@@ -289,6 +289,7 @@ public class CraftingPattern implements ICraftingPattern {
     			itemHashCode = mItemStack.getCount();
     			itemHashCode = itemHashCode * 31 + mItemStack.getItem().hashCode();
     			itemHashCode = itemHashCode * 31 + mItemStack.getItemDamage();
+    			itemHashCode = itemHashCode * 31 + Objects.hashCode(mItemStack.getTagCompound());
     			hashCodeCached = hashCodeCached * 31 + itemHashCode;
     		}
     	}
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
@@ -28,7 +28,8 @@ public class CraftingPattern implements ICraftingPattern {
     private List<List<ItemStack>> oreInputs = new ArrayList<>();
     private List<ItemStack> outputs = new ArrayList<>();
     private List<ItemStack> byproducts = new ArrayList<>();
-
+    private Integer hashCodeCached = null;
+    
     public CraftingPattern(World world, ICraftingPatternContainer container, ItemStack stack) {
         this.container = container;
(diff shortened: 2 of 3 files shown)


src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPattern.java
```diff
@@ -14,7 +14,10 @@
      * @return the {@link ICraftingPatternContainer} where the pattern is in
      */
     ICraftingPatternContainer getContainer();
-
+    
+    public int hashCode();
+    
+    public boolean equals (Object obj);
```
raoulvdberge: These methods do not belong on an interface.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +273,24 @@ public String toString() {
...
+    String toHash;
+    @Override
+    public int hashCode() {
+    	if(toHash==null){
+    		toHash=new String();
+    		for(ItemStack mItemStack : this.getOutputs()){
+    			toHash+=mItemStack.toString()+" ";
+    		}
+    	}
+    	return toHash.hashCode();
```
raoulvdberge: This is a bad way of implementing a hashcode. Use something else than strings.

raoulvdberge (changes requested):
Please fix your coding style to match the RS code.

raoulvdberge (approved):

raoulvdberge:
Please fix your coding style to match the RS code.

raoulvdberge (changes requested):
Please fix your coding style to match the RS code.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +273,28 @@ public String toString() {
...
             '}';
     }
-
+    
+    @Override
+    public boolean equals (Object obj) {
+    	return this==obj || (obj instanceof ICraftingPattern && this.alike((ICraftingPattern)obj));
+    }
+    
+    Integer hashCode;
```
way2muchnoise: Would mark this private and any reason for it being the boxed `Integer` instead of the primitive `int`?

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +273,28 @@ public String toString() {
...
+    @Override
+    public int hashCode() {
+    	if(hashCode==null){
+    		hashCode=0;
+    		for(ItemStack mItemStack : this.getOutputs()){
+    			int itemHashCode=0;
+    			itemHashCode=mItemStack.getCount();
+    			itemHashCode=itemHashCode*31+mItemStack.getItem().hashCode();
+    			itemHashCode=itemHashCode*31+mItemStack.getItemDamage();
```
way2muchnoise: Might want to include the hash of the NBT data, since a lot of modded items use that as an identifier.

way2muchnoise (commented):
Looks pretty good, added some small comment here and there.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
raoulvdberge: I think he used boxing for caching the hash (if null compute hash).

Either way, this field has to be moved to the top like other fields and indeed needs a visibility modifier (private).

way2muchnoise: Might want to make it `Integer hashCode = null;` and maybe add a little like saying this is for caching reasons. Or just call it `hashCodeCached`

raoulvdberge (approved):

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +274,28 @@ public String toString() {
...
+        if (hashCodeCached == null) {
+            hashCodeCached = 0;
+            for (ItemStack mItemStack : this.getOutputs()) {
+                int itemHashCode = 0;
+                itemHashCode = mItemStack.getCount();
+                itemHashCode = itemHashCode * 31 + mItemStack.getItem().hashCode();
+                itemHashCode = itemHashCode * 31 + mItemStack.getItemDamage();
+                itemHashCode = itemHashCode * 31 + Objects.hashCode(mItemStack.getTagCompound());
+                hashCodeCached = hashCodeCached * 31 + itemHashCode;
+                }
```
raoulvdberge: Wrong indentation of braces.....

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +274,28 @@ public String toString() {
...
+    @Override
+    public boolean equals (Object obj) {
+        return this == obj || (obj instanceof ICraftingPattern && this.alike((ICraftingPattern)obj));
+    }
+    
+    @Override
+    public int hashCode() {
+        if (hashCodeCached == null) {
+            hashCodeCached = 0;
+            for (ItemStack mItemStack : this.getOutputs()) {
```
raoulvdberge: Bad variable name "mItemStack"

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
```diff
@@ -273,7 +274,28 @@ public String toString() {
             ", byproducts=" + byproducts +
             '}';
     }
-
+    
+    @Override
+    public boolean equals (Object obj) {
+        return this == obj || (obj instanceof ICraftingPattern && this.alike((ICraftingPattern)obj));
```
raoulvdberge: Still wrong on the parens. They require a space, so `(ICraftingPattern) obj`.

raoulvdberge:
Thanks!

PR #1610, opened 2018-01-09
Fixed issue with pattern overwriting

ineternet:
Pattern overwriting if the blank pattern (top slot in the pattern slots) slot was empty. Fixed this by only considering a recipe invalid if there are *absolutely no* patterns available, meaning none in top and none in bottom.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/node/NetworkNodeGrid.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/node/NetworkNodeGrid.java
@@ -455,7 +455,8 @@ public class NetworkNodeGrid extends NetworkNode implements IGrid {
     }
 
     public void onCreatePattern() {
-        if (canCreatePattern()) {
+        if (
+        ) {
             if (patterns.getStackInSlot(1).isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/node/NetworkNodeGrid.java
```diff
@@ -491,9 +491,13 @@ public void onCreatePattern() {
             patterns.setStackInSlot(1, pattern);
         }
     }
+    
+    private boolean patternAvailable() {
```
raoulvdberge: `patternAvailable` is a weird name for a method that checks if two pattern slots AREN'T available...

ineternet: its supposed to return the opposite, sorry! commited that change

raoulvdberge:
Can you rename the method to add `is` in front of it. So `isPatternAvailable`.

Also, can you add this to the changelog in CHANGELOG.md (in the existing format)?

raoulvdberge:
Awesome! Thank you for this PR.

Issue #1604, opened 2018-01-06
Craftable items are counted as itemcount 1


raoulvdberge:
![x](https://drive.google.com/uc?id=1l9K_9ysO3mscWoLELu9WJOKWmqHvA6OM)

raoulvdberge:
Fixed by #1613.

PR #1613, opened 2018-01-09
Changed item quantity of a craftable item to 0

ineternet:
Fix for #1604 
Whenever a system has none of a specific item but the ability to craft it, the grid saves the quantity as 1. Quantity sorting would make use of this and sort it alphabetically together with other quantity 1 items. This change should make quantity sorting think craftable items have a quantity of 0, making all of them appear behind itemstacks with the quantity 1 (or higher).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 1.5.32
 - Added Spanish translation (Samuelrock)
 - Fixed issue where the Pattern Grid can only overwrite patterns when blank ones are present (ineternet)
+- Changed stack quantity of craftable items from 1 to 0 to fix Quantity Sorting (ineternet)
 
 ### 1.5.31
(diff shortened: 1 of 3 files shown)


raoulvdberge:
Can't you change `getQuantity` in `GridStackItem` instead? That should be a little cleaner.

ineternet:
`GridStackItem` should now still store craftable items as 1, but tell any other class they're empty. This makes GridSortingQuantity think craftable itemstacks have less items on them than actual single items in the system. Although I'm unsure if any other classes will have a problem with the itemstack being supposedly empty?

raoulvdberge:
Shouldn't cause any issues. These are the only usages of the `getQuantity` method:
x

raoulvdberge:
Can you add a changelog entry again?

raoulvdberge:
Thank you.

PR #1630, opened 2018-01-19
Stop fluid stacks from displaying "0"

ineternet:
Could be confusing. Instead they will display "<1", to indicate the stack isnt empty but contains less than a bucket.


raoulvdberge:
Can you send me a screenshot of how this looks?

PR #1641, opened 2018-01-27
Keep modes to search box and shift+click for previous mode

CJBok:
Added Keep modes to search box:
- Normal (keep)
- Normal (keep, autoselected)
- JEI synchronized (keep)
- JEI synchronized (keep, autoselected)

Keep mode will load your previous search text in the search box when opening a Grid GUI.

Added Shift+Click on search box mode button to go to previous mode.


way2muchnoise:
Might be a better idea to have a single flag for "saving" the search. Also using the `static String` will make it so that every Gird you open will have that search. An other approach would be to save it to the TE.

Also right-click on the buttons might also be a good idea for going to the previous mode. But shift+click works.

raoulvdberge:
I don't think I like this idea.

1) It adds a lot of spam to the mode toggle button for just 1 extra mode
2) The same can be achieved with the grid history feature (only 1 button press "up" required)

raoulvdberge:
I closed this because I don't like the idea. Thank you anyway!

Issue #1643, opened 2018-01-27
Weird NBT compare bug in interface


raoulvdberge:
https://www.youtube.com/watch?v=vhvkYKTex_o&feature=youtu.be

ineternet:
Probably is caused by the export mechanism not checking NBT and thinking "oh hey, i dont need to export the item anymore, put it back into the system".

PR #1644, opened 2018-01-27
Interface NBT/Metadata check fix

ineternet:
Fix for #1643 
Tested this with the exact case from the provided video (two items, different NBT tags). Reproduced same result, this change gave expected result.

oh also, this includes changes from my previous PR so sort that one out first before accepting this one


raoulvdberge:
You should've done this fix on a separate branch, now, you're having conflicts and I can't merge.

ineternet:
closed in favor of #1646

Issue #1147, opened 2017-04-09
Only the topmost disk in the disk manipulator is manipulated


ghost:
**Minecraft version**: 1.10.2
**Refined Storage version**: 1.2.25
**Forge version**: 12.18.3.2272

way2muchnoise:
Intended behaviour, once that one is done it should do the next one.

PapiDimmi:
Firstly, this is not a question.

Secondly, the disk manipulator only affects the topmost disk, like I said. It doesn’t move on to the other disks once it’s done.

I _assume_ that it only moves on to the next disks if the disks in the input slot are placed in the output slot, but that only happens if the disks are completely empty, which they aren’t in my case.

way2muchnoise:
And you are sure that the items on the drive can be moved to your system?

PapiDimmi:
Yes. Whenever I use the disk manipulator, I use it to move some items from my disks to my storage drawers, but unless the disks go in one of the output slots, which they only do if they’re completely empty, the disk manipulator doesn’t move on to the next disk.

dovahkiin4991:
just get another disk you use when you are redumping items, that way it will empty as much of it into the drawers as possible and than any leftovers will just go to the new disk leaving you with an empty disk for next time you do this. Also making sure the priority is set to make sure everything possible will go to the drawers.

raoulvdberge:
Intended behavior.

PR #1651, opened 2018-01-30
Fix for #1147

ineternet:
Read commit for explanation.

Tested before+after, could reproduce unintended (before fix) and intended (after fix) results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Changed stack quantity of craftable items from 1 to 0 to fix Quantity Sorting (ineternet)
 - Changed fluid stack amount to not display "0" anymore (ineternet)
 - Fixed NBT/metadata check on exporting in an Interface (ineternet)
+- Fixed Disk Manipulator being stuck on unemptiable, non-empty disks (ineternet)
 
 ### 1.5.31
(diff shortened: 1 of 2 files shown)


raoulvdberge:
I'm not sure if I like the current behavior more. It's more predictable (as an end-user) to say "is the disk empty" then to assume "are the disk contents all insertable"?

@way2muchnoise, what do you think?

way2muchnoise:
Both behaviours make sense to me, mostly personal preference imo. So all up to you @raoulvdberge ;)

ineternet:
I agree that it might be more intuitive to assume that only an empty disk is a fully transferred disk. This is really based on personal preference, whether you want to know if it's done transferring or "waiting for free space".

In that case, assuming you want to keep the old design, why's the issue still open? This PR is all I could think about when I read the issue.

ineternet:
I've changed the behaviour.

The disk now does *not* get moved to an output slot when it cannot be transferred, but is instead skipped, meaning that the next possible disk (second, third slots) is starting to import. As soon as there's space for more from the first disk, it stops and prioritizes the first disk again, until it is empty (moved to output) or untransferable (skipped over, continuing with second/third disk).

raoulvdberge:
Thanks!

PR #1665, opened 2018-02-11
Fix orientations for Portable Grid

TeamSpen210:
In hand currently, the orientation is rather nonsensical. This adds the `"display"` section to the [https://minecraft.gamepedia.com/Model#Item_models](model JSON) to give poses for first/third person, and eliminates some rendering code which seems to be a workaround to implement this for the `"gui"` pose (in inventory).

I wasn't quite sure how to hold this (given that it's rather bulky), but decided to rotate it sideways to use more like a tablet. It might work a bit better if the item used an alt model with the screen tilted flat.

Here's a pic of the poses:
port_grid_poses


raoulvdberge:
I don't like this, the current rotation is fine. It's supposed to look like a laptop in hand (and in world, like it does currently), and making it upside down is rather weird.

Thanks anyway.

TeamSpen210:
Curently it doesn't, especially in the left hand - it's floating out off to the side.

raoulvdberge:
Sorry, that is correct indeed.. My mistake.

Can you make a new PR fixing this and making it horizontal instead of vertical? Thanks!

PR #1675, opened 2018-02-17
Fix orientations for Portable Grid, v2

TeamSpen210:
Now with a more laptop-like holding position.

port

JSON also has the advantage that users can override with resource packs, if they want to for whatever reason.


raoulvdberge:
Thank you, this is great! :)

PR #1740, opened 2018-04-08
Allow crafters to be daisy-chained

tomKPZ:
In crafting situations that require more than the 9 blocking patterns
that one crafter allows, the only solution is to duplicate the setup
to allow a second crafter.  This CL allows crafters to be
daisy-chained to provide a solution to this issue.

Also, since unlimited crafters can now be placed on a single
inventory, this eliminates the need to have multiple intermediate
inventories which then pipe into the desired inventory.  This also
improves the crafting manager UI as now those patterns can all be
listed under one "Furnace" instead of (potentially multiple) "Chest"s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Refined Storage Changelog
 
+### 1.5.34
+- Allow crafters to be daisy-chained (tomKPZ)
+
 ### 1.5.33
 - Added Crafter Manager (raoulvdberge)
 - Patterns in the Crafter slots now automatically render the output without pressing shift (raoulvdberge)
@@ -10,7 +13,6 @@
 - Fixed Grid not always using all combinations when using JEI autocompletion (raoulvdberge)
 - Increased Grid performance (raoulvdberge)
 - Various internal refactors (raoulvdberge)
-- Allow crafters to be daisy-chained (tomKPZ)
 
 ### 1.5.32
 - Added Spanish translation (Samuelrock)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed Grid not always using all combinations when using JEI autocompletion (raoulvdberge)
 - Increased Grid performance (raoulvdberge)
 - Various internal refactors (raoulvdberge)
+- Allow crafters to be daisy-chained (tomKPZ)
 
 ### 1.5.32
 - Added Spanish translation (Samuelrock)
--- a/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPatternContainer.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPatternContainer.java
@@ -8,6 +8,8 @@ import net.minecraftforge.items.IItemHandlerModifiable;
 
 import javax.annotation.Nullable;
 
+import com.raoulvdberge.refinedstorage.api.network.INetwork;
+
 import java.util.List;
 import java.util.UUID;
 
@@ -82,9 +84,10 @@ public interface ICraftingPatternContainer {
     boolean isBlocked();
 
     /**
+     * @param network the network associated with this container
      * @param blocked whether the container should be blocked
      */
-    void setBlocked(boolean blocked);
+    void setBlocked(INetwork network, boolean blocked);
 
     /**
      * @param blockedOn the descendant that this container is blocked on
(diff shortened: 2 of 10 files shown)


way2muchnoise:
Could you add a picture or maybe a small video of how this exactly looks in world?

tomKPZ:
Here's a demo:
https://www.youtube.com/watch?v=cH5ITjsbCc8

You might have noticed a bug at the end of the video where the crafter manager showed the two crafters I added under "Crafter" rather than "Chest".  I was briefly confused :)  However, this was easily fixed in the latest revision by rebuilding the crafting manager when a crafter is rotated.

raoulvdberge:
You're still making lots of changes, so if you believe this is bug free and ready for review, let me know.

tanderson-google:
Should be ready at this point

tomKPZ:
Should be ready at this point

Saklad5:
This is a spectacularly good idea. There’s a lot of automation ideas I’ve decided against purely because I’d have to replicate the build to add more than nine recipes.

Saklad5:
And I hate adding unnecessary steps like a buffer chest. Even something as simple as a chest can make failures more likely.

It’d be great if Crafters could output directly to pipes as well, but that might be a little finicky to implement.

tomKPZ:
friendly ping :)

raoulvdberge:
I am going to release a new version soon, and this PR will be merged *after* that.

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingManager.java
```diff
@@ -67,6 +76,30 @@ public CraftingManager(TileController network) {
         return containerInventories;
     }
 
+    @Override
+    public void addContainerBlock(UUID blocker, UUID blockee) {
```
raoulvdberge: Why is it necessary to store `blocker`?

src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/node/NetworkNodeCrafter.java
```diff
@@ -222,15 +253,48 @@ public boolean hasConnectivityState() {
...
     public boolean isBlocked() {
-        return blocked;
+        ICraftingPatternContainer proxy = getProxyPatternContainer();
+        return proxy != null && network != null && network.getCraftingManager().isContainerBlocked(proxy.getUuid());
     }
 
     @Override
-    public void setBlocked(boolean blocked) {
-        this.blocked = blocked;
+    public void setBlocked(INetwork network, boolean blocked) {
```
raoulvdberge: With the extra `network` parameter, and this method having no side-effects for `NetworkNodeCrafter` anymore, I feel like this method can be removed and moved to `CraftingManager`.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Fixed Grid not always using all combinations when using JEI autocompletion (raoulvdberge)
 - Increased Grid performance (raoulvdberge)
 - Various internal refactors (raoulvdberge)
+- Allow crafters to be daisy-chained (tomKPZ)
```
raoulvdberge: Can you move this entry to a new 1.5.34 list in the changelog?

src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPatternContainer.java
```diff
@@ -53,12 +68,29 @@
     BlockPos getPosition();
 
     /**
-     * @return true if this container is blocked, false otherwise
+     * Containers may be daisy-chained together.  If this container points to
+     * another one, gets the last container in the chain.  If containers are
```
raoulvdberge: last container? Don't you mean first container? The root one (the one at front) is the first one.

src/main/java/com/raoulvdberge/refinedstorage/api/autocrafting/ICraftingPatternContainer.java
```diff
@@ -53,12 +68,29 @@
...
+     * not daisy-chained, returns this container.  If there was a container
+     * loop, returns null.
+     *
+     * @return the root pattern container
+     */
+    @Nullable
+    ICraftingPatternContainer getProxyPatternContainer();
```
raoulvdberge: I'd call this one `getRootContainer`.

raoulvdberge (changes requested):

tomKPZ:
Did all of the things.

raoulvdberge:
Thank you, great PR!

raoulvdberge (approved):

Issue #1762, opened 2018-04-25
[Feature request] Search boxes should handle repeated key events


tomKPZ:
#### Steps to reproduce:
1. Hold down the backspace key (or any other key) in the search box.

#### What happens:
Only one character gets deleted.

#### What you expected to happen:
After a short time, multiple characters should start being deleted.

#### Version (make sure you are on the latest version before reporting):
Minecraft: 1.2.12
Forge: ToT
Refined Storage: 1.5.34

raoulvdberge:
This works for me:

![x](https://drive.google.com/uc?id=1mcThdLnZylPpyGOChvStUue6ZEdWTMF)

tomKPZ:
Hm.. when I build RS and run it from gradle, I can't repro the issue.

However, I can reliably repro on the Direwolf20 1.12.2 pack with an updated RS and forge.  It also repros with the standard RS and forge that ship with the pack, and when all mods except RS are disabled.

tomKPZ:
I figured out how to repro it from my gradle client:
1. Load up a world in creative mode with JEI cheating enabled
2. Set JEI cheating to false
3. Try to type repeated characters in the pattern grid

raoulvdberge:
Ah, I assume that's a flag that's by default on, but anyone can toggle it and it won't be untoggled.

So I need to assume that it'll be always untoggled and toggle the repeated key events on in any case.

PR #1784, opened 2018-05-12
Fix repeated key events not getting handled in search boxes

tomKPZ:
GUIs are supposed to call enableRepeatEvents() from initGui() and
onGuiClosed() to properly handle repeated key events.  This change
adds the appropriate calls in GuiBase.

See the equivalent code in eg. Applied Energistics 2:
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/cc9b33b47335ddb7be29a007de3f7e364fdc1859/src/main/java/appeng/client/gui/implementations/GuiMEMonitorable.java#L224
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/cc9b33b47335ddb7be29a007de3f7e364fdc1859/src/main/java/appeng/client/gui/implementations/GuiMEMonitorable.java#L399

Fixes https://github.com/raoulvdberge/refinedstorage/issues/1762


raoulvdberge:
Thank you!

PR #1835, opened 2018-06-27
Fixed a minor problem with redudant calls to `world.getBlockState(pos)`

samtrion:
Fixed a minor problem with redudant calls to `world.getBlockState(pos)`


raoulvdberge:
Undo the capability cast changes, these are necessary to avoid warnings.

This also doesn't really warrant a changelog entry, because these weren't actual perf issues in the first place ;)

src/main/java/com/raoulvdberge/refinedstorage/item/CapabilityProviderEnergy.java
```diff
@@ -11,12 +11,10 @@
 import javax.annotation.Nullable;
 
 public class CapabilityProviderEnergy implements ICapabilityProvider {
```
raoulvdberge: The changes in this file are dangerous, and involve capability creation semantics. Please undo.

raoulvdberge:
Thank you!

Samtrion:
you are welcome

PR #2076, opened 2018-11-04
Add a hard-dep on Forge 2694

josephcsible:
As of commit bfaae24, we'll crash on startup with earlier versions of Forge.


raoulvdberge:
Which line is causing the problem?

PR #2125, opened 2018-12-04
Changes block storage and block fluid storage to whitelist by holding wrench

bretterer:
Just an idea here to change the block storages to be whitelist by default. The purpose of this change is if you currently have automation of fluids on your system, and you want to add a new block fluid storage or block storage for items, it will quickly begin to fill these blocks. 
To get around this, you have to disconnect your blocks to change to a whitelist and then re-connect them.

This PR addressed this by setting the mode to `WHITELIST` on the block when placing it if you have the wrench in your offhand so you can place it in the world connected to your RS system without disconnecting everything for a second while you switch it to whitelist.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Mon Dec 03 18:17:39 EST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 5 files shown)


raoulvdberge:
No, this is too hidden. Users won't know this feature exists and there is no way to make it clear. Additionally, I don't like changing the placement mechanics in odd ways like this. 

Thx either way

bretterer:
No worries, I was just following the same idea that storage drawers does to place locked drawers in the world

bretterer:
@raoulvdberge coming back to this,. Can we at least change the blocks to be whitelist by default?  When adding a block to an existing storage system, I'd love to have it not start filling by default for fluids I auto store

raoulvdberge:
Nope. Too many people will come asking why their storage doesn't work.

PR #2137, opened 2018-12-13
updated german translation

Cydhra:
Updated the german translation with the new entries added since the last time


raoulvdberge:
Thx. I will take care of this next month.

PR #2152, opened 2019-01-03
fix: set count for matched oredict results

the-eater:
Currently when an oredict processing pattern matched a different input that wasn't in the original pattern, (say a copper ignot from IC² instead of Thermal Expansion) it doesn't set the count, and it will try to use only 1, as seen in the screenshots

pattern
now_oredict
wat_is_this

With this fix:

i_did_that

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/CraftingPattern.java
@@ -70,6 +70,11 @@ public class CraftingPattern implements ICraftingPattern {
                         }
                     }
 
+                    // Fix item count
+                    for (ItemStack ore: ores) {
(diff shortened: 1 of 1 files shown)


the-eater:
any indication when this later may be? :)

raoulvdberge:
No, sorry

the-eater:
Ah that's cool, then I'll just run on my fork for a while, thanks for your work!

raoulvdberge:
Would it be possible to reference some issues on the issue tracker that are related to this fix?

the-eater:
Actually have not seen any issues that seem to be caused to this, the one's that may be are to vague to make that conclusion.

The reason why I think it's not widespread is because it's only a pattern issue, and only applies when an actual stack size is specified which is pretty uncommon, if you want I can create a simple test world for you which shows you the issue. (although the code flow should already display what's going wrong)

raoulvdberge:
Thx!

PR #2158, opened 2019-01-06
Let constructor & destructor interact with world using their owner's profile. + Fixed CraftingTask duplication bug

Radviger:


raoulvdberge:
1) Regarding the constructor/destructor change: what does this change in practice?
2) Regarding the CraftingTask dupe bug
    - Would it be possible to reference some issues on the issue tracker that are related to this fix?
    - What exactly does it fix?
    - How can I reproduce it myself?

Radviger:
1. It allows them to place/break blocks as if they were placed/broken by the owner of that particular constructor/destructor. It's almost impossible to play on servers with region protection mods/plugins without this feature because it isn't really safe to grant region permissions to the default '[Minecraft]' profile.

Radviger:
2. Here's the issue with a comprehensive explanations: github.com/raoulvdberge/refinedstorage/issues/2143

Radviger:
Looks like a bug in github mobile version that screwed up it's display bame. Link seems broken but I can click on it and it works well for me.

raoulvdberge:
Thanks!

Issue #2088, opened 2018-11-13
Grid does not update Patterns without opening and closing it.


Darkere:
#### Issue description:
When a grid is open and a pattern is inserted (via automation) it is necessary to close and reopen the grid to be able to see it.

#### Steps to reproduce:
1. Create a hopper line that inserts into a crafter.
2. Insert a pattern into the hopper and wait for the pattern to appear in the grid.
3.
...

#### Version (make sure you are on the latest version before reporting):

- Minecraft:1.12
- Forge:2772
- Refined Storage:1.6.9

#### Extra Info
IIRC inserting patterns via automation should not be possible because of security? 
It would be nice to see this fixed regardless as the Reborn Storage Multiblock does have the ability for insertion via automation.

raoulvdberge:
This is really minor and doesn't really matter.

Darkere:
It allows for a really nice setup where instead of searching for a recipe, making the pattern, inserting the pattern and then searching for the item again, you can instead throw the item on the ground, have it collected and inserted via automation. 

It saves on typing out the search twice. (I know about the history feature but the down button is nowhere close to where my hands are when playing..).

PR #2233, opened 2019-05-10
Crafters update open grids if they receive a new pattern implements #2088

Darkere:
I would prefer a cleaner way of doing things but I don't want to break the API for such a minor change.


raoulvdberge:
Hey,

I decided not to merge this since IMO there are too much changes to core classes that I don't like, for such a small feature.

If there is a cleaner way to do this without changing all those important classes, I'm interested in merging a new PR.

PR #2238, opened 2019-05-13
limited craftingmonitor updatespeed depending on tasksize fixes #2132

### LINKED ISSUE #2132: Performance issue with Crafting Monitor
Not 100% sure what exactly triggers this as this doesn't always happen but it seems to be a combination of a large "ingredient" list and fast changing of stored and crafting values.

refinedstorage-1.6.11

Darkere:
numbers may need some tweaking
currently set to:
-update on change for 0-1000
-update every tick for 1000-10000
-wait for an additional tick up to 20 for every 10000 lookups

The actually expensive part of the calculation is summing up the outputs of all the recipe outputs in the queue. It might be possible to multithread this part for better performance. 

I've not looked into other potential fixes that modify the current crafting procedure.


Darkere:
Wait on this one if the autocrafting changes I want to make work this is no necessary.

raoulvdberge:
I'm closing this in favor of your autocrafting PRs

Issue #1790, opened 2018-05-18
Payload Max Size Crash/Disconnect


artdude543:
#### Issue description:
A maker of SevTech: Ages here! Got an issue posted on our tracker about the usage of BWM Recipes on RS Pattern. Full details about the issue and how to reproduce it can be found on our ticket: https://github.com/DarkPacks/SevTech-Ages/issues/2498 _(Saves on copy/paste)_

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.12.2
- Forge: 14.23.3.2690
- Refined Storage: 1.5.33
- BWM (Any version shows this issue)

Does this issue occur on a server? [yes/no] yes

#### If a (crash)log is relevant for this issue, link it here:

https://gist.github.com/nitrous24/01f1ff00a7c3d05df2aeab46744f24e3

instinxx:
Have same issue that results in "Payload may not be larger than 32767 bytes". The player shift clicked the + sign in JEI to build chests in the crafting grid.
Forge Mod Loader version 14.23.4.2760

CrashLog here: https://pastebin.com/f8pSZ2p7

Modlist Here: https://pastebin.com/gGTwhB2i

valveo:
https://github.com/CoFH/Feedback/issues/865
This is same issue.

raoulvdberge:
It's not.

valveo:
Sorry, corrected link. Its related forestry bug. There are so many saplings, wood types that causing issues.

Yamza:
Seems like this may be related: https://github.com/AllTheMods/ATM3-Remix/issues/82

Mitsuma:
Any progress on this one?
Today playing ATM3 I encountered a similar issue with the same payload error message.

I just tried to use the + from a regular chest into the pattern computer and then it disconnects. 
Updated Refined Storage to 1.6.12, still getting the same issue, SP or Server.

SuperSandro2000:
This issue is almost a year old by now. Does this need to be fixed on JEI side or RS or both?

turret001:
I don't know if it's RS or JEI but it still exists and it's quite annoying

jlsTyrael:
[...]

PR #2242, opened 2019-05-16
catch payload too large for #1790

Darkere:
So from what I've gathered we ideally would want JEI to tell us the Registry name of the recipe so that could be sent instead. I can't find that though so until then just not making it crash is at least something we can do.

As for what to actually do after catching the error. I could:

1. Recursively split the package in half and send multiple.
2. Render something by adding a new value into the container and reading that from the GUI.
3. return a JEITransferError. The convenient User facing errors don't work at that point anymore though. But this makes clicking the button do nothing.
4. Just keep what it does now which makes it look like it worked but doesn't actually do anything.

Kinda hoping there is a much cleaner solution somewhere so if anyone has any suggestions....


raoulvdberge:
We can't use the registry name since processing recipes don't have any registry things set.

Why are you manually throwing exceptions if you made a custom buffer?

Darkere:
I couldn't figure out a way to stop sending the package from inside the Message. In hindsight, it's probably a better idea to construct the buffer before creating the message.

Darkere:
So this solution is terrible. I don't even need to make the custom buffer if I want the size...

Gonna try to catch it before the message by counting the item amounts.

Issue #1937, opened 2018-08-03
Insert order for processing patterns


ghost:
#### Issue description:
Some machines with multiple slots, e.g. the Slice'n'splice from EnderIO, required that the items are inserted in the correct order. Currently there seems to be no way to force the insertion order, so the crafting does not work.

#### What happens:
The pattern for a Skeletal Contractor requires the silicon to be in the middle slot in the bottom, but the silicon is inserted first in to the bottom left slot. The rotten flesh and skeleton head are not accepted.

#### What you expected to happen:
A way to force the insertion order so that rotten flesh is inserted first.

#### Steps to reproduce:
1. Have a slice'n'splice with a crafter
2. Create a pattern for a skeletal contractor, how the items are placed in the crafting grid does not matter.
3. Request a craft of the skeletal contractor.
4. Look in the slice'n'splice, the items are not in the correct slots.

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.12.2
- Forge: 14.23.4.2739
- Refined Storage: 1.6

Does this issue occur on a server? [yes/no]
A local LAN server, yes.

eihns:
Maybe just add the items in the order you put it in the crafting field...

Like:

1 2 3 
4 5 6 
7 8 9

So could just insert the first item in the upper left slot, then the second... i guess u get it. Would be an easy fix, i guess?

DarkHelmet433:
I've had this problem with this machine in the past.  I used to solve it by changing the order of the pattern so that the head always came first.  Once the head was inserted, the remaining slots would only accept the correct item.

However in 1.6+ and the new EnderIO, I've literally used the recipe exactly as it's specfied in JEI.
I took a screenshot from my world showing this: https://i.imgur.com/X919G4S.png
In the past I've had to have the head in the top left slot of the pattern.

This machine is quirky at best.

eihns:
The problems continue... alloy smelter... dark steel...

u need crushed coal, iron and obsidian... it first inserts crushed coal and iron... which gives steel...

changhe3:
This would also be very useful for autocrafting using things like AA Empowerer.

MarkusPriska:
Would also really like this feature as i kinda need it for automatic crafting

starryknight64:
I was able to workaround this issue by hooking the crafter up to a chest and have the Slice'n'Splice accept input from the chest.

Norway174:
[...]

PR #2246, opened 2019-05-21
Autocrafting rework

Darkere:
Part 1, Processing.
It rewrites processing to iterate over each pattern instead of each processing step. 
This makes requests that have many of the same items much faster. 
Before: https://sparkprofiler.github.io/#GWZTTP4QkF
After: https://sparkprofiler.github.io/#6YRrjGQjZH

Things that this will change for users:
- ~~Processing tasks that still have items that are in processing will no longer display "Cannot Insert" for items that have not been inserted yet~~ reverted its annoying but you really only need to look at the crafting monitor when something is not working. In that case, it's needed.
- ~~processings in locked/no machine/cannot insert state will block all crafters with the same pattern (this is fixable but not a priority)~~ fixed
- ~~Insertion Speed is much lower. This is because `getMaximumSuccessfulcraftingUpdates()` is being ignored. I'll leave the decision up to you whether you want to call insertion X times or just insert X * items at once.~~ Implemented X times as performance seems fine.

I tested in a world with a fairly big auto crafting setup and it worked just fine, but considering the number of bugs I produced and fixed while making this and the number of times I had to redo parts of it, bugs are (likely) still present.
Also not happy with many of the method names. Taking suggestions. 

Surprisingly enough the current Implementation works fine with 1.6.14 crafting tasks. They just stop and can be canceled. 
I expect this will change once I get to Part 2 & 3 Crafting and Calculation.


Darkere:
Alright, Part 2 is done.
Same as part 1 just for crafting. 
Crafting 2 6xCobble + 64 ultimate induction cells (with max size reborn storage crafter) and watching in the crafting monitor:  https://sparkprofiler.github.io/#2CyaHsCMCR

I've also fixed 3 bugs from the previous commit and implemented inserting X times. This should bring insertion Speed back up what it was before. 

Calculation might be trivial but I want to understand how it works before actually implementing anything.

Darkere:
Alright, done (finally).

This adjusts the crafting calculation to work with quantities so that multiple crafts can be created at once.
That, in turn, makes adding all ingredients together possible. For Chests, for example, it will now request 8 Wood instead of requesting wood 8 times.
There are faster ways to count the number of occurrences in a set than what I implemented, but I failed at making them work in this case...

Things like crafting octuple compressed cobble now totally possible:
https://gfycat.com/FarawayCloseArmadillo
Although crafting more than 49 ends up in an Integer overflow...

Also fixed some math from previous commits.

I still need to do more testing to confirm everything working. But I've not found anything broken so far.

raoulvdberge:
Very interesting! This won't be merged soon though. I'm going to take my time to take a GOOD look at this code, to make sure it doesn't break anything and I understand it myself.

(I'm avoiding making the same mistake as I did last time - merging autocrafting changes without really understanding what changed. This lead to me not being able to maintain the code myself anymore and having to rewrite it all. So that's why I am going to take my time investigating this.)

Darkere:
I was hoping you would. I am sure there are some bugs left. 

You should already be familiar with the majority of the code as much of it is the same or close to the same as before.

bigenergy:
@raoulvdberge  Can you find out when you can tackle this problem? Thanks for the answer!

Darkere:
There is still a major part (oredict) missing. I have it mostly figured out but not bugfree just yet.

Darkere:
Alright finally got oredict working. 

Shoot me a message on discord for any questions about my (questionable) decisions. (Or ask here I guess..)

raoulvdberge:
I'll look into this for the release after the next one Darkere, gonna first do some quick/easy fixes.

Darkere:
This will not get merged for 1.12 as 1.12 will only get smaller fixes at this point. 

#2329 is a small change that fixes the most significant problem without reworking the entire crafting logic and may be merged for 1.12 in the future.

The changes from this Pull Request will get another rework and may be merged for 1.14.

raoulvdberge:
I never said I wasn't going to merge this for 1.12, I just need some time for after 1.14.

Darkere:
Summary of Changes:
Sorting of both the Crafting Monitor Display and the Crafting preview
implemented insert order for Patterns #1937
Due to this change, these to Recipes are no longer the same: 
image

In fact, the right recipe will not work at all as the machine itself only has 2 slots and there will be no attempt at inserting a combined set. 

improved Performance for: 
- Crafting Calculation
- craft updating
- NBT read and write
- item Importing while crafting tasks are running

This contains API changes. Pretty sure no one was using that part of the API but...

raoulvdberge:
Due to the size of this PR this won't be pulled for mc 1.14.

PR #2256, opened 2019-06-08
Shorted text. To fix the overlays.

yaroslav4167:
GUI without this fix:
jP1i


src/main/resources/assets/refinedstorage/lang/ru_ru.lang
```diff
@@ -7,11 +7,11 @@ key.refinedstorage.clearGridCraftingMatrix=Очистить сетку созд
 gui.refinedstorage:controller.0=Контроллер
 gui.refinedstorage:controller.1=Творческий контроллер
 gui.refinedstorage:grid=Терминал
-gui.refinedstorage:grid.craft=Создание
-gui.refinedstorage:crafting_grid=Терминал создания
-gui.refinedstorage:pattern_grid=Терминал шаблонов
+gui.refinedstorage:grid.craft=+
```
CrossVas: It's a good idea

CrossVas:
It's not a good way to fix this problem.
Instead of making the text shorter, it's better to make Search bar to be 4 slots long (if this is possible)

yaroslav4167:
Shorten terminal name displays only in terminals GUI.
I think it is not difficult to distinguish the terminals inside.

raoulvdberge:
I think it's a fine fix.

raoulvdberge:
I will merge it later.

PR #2269, opened 2019-06-21
Fix Oredict redefining recipes fixes #2216

### LINKED ISSUE #2216: Certain Processing Oredict patterns will redefine itemstacks in all mods
#### Issue description:
When setting certain items (so far only gold ingots and iron ingots) to use processing oredict patterns with 2+ items in a single input slot, the itemstack will be redefined in all recipes, including modded recipes to use the same quantity. Regular crafting works okay in spite of this, but modded recipes that use the same input item will be re-defined to use whatever the quantity is in the pattern input which essentially breaks them, so this is not just a cosmetic inconvenience in JEI. **This bug will persist even after restarting the game** unless the pattern is changed.

#### What happens:
See above

The item quantity will not be re-defined in all recipes

#### Steps to reproduce:
(Updated modpack instance with **only** most recent versions of RS, JEI, and Forge, no other mods)
1. Place down a controller (creative or regular + powered)
2. Place an adjacent pattern grid
3. Create a oredict processing pattern with 2+ gold or iron ingots in an input slot and anything in the output slot. Check JEI uses for that input item.
4. Re-defining the pattern to only use 1 input in a slot fixes this bug. Restarting the game without changing the pattern **does not fix this**

Curiously enough, I have only been able to reproduce this issue with gold or iron ingots as inputs so far.

[DEMONSTRATION](https://gfycat.com/hairyhighlevelazurewingedmagpie)

- Minecraft: 1.12.2
- Forge: 14.23.5.2836
- Refined Storage: 1.6.14
- JEI: 1.12.2-4.15.0.276

Unsure, only tested in single player

Darkere:


raoulvdberge:
Thank you

PR #2285, opened 2019-07-21
restrict gregtech rendering ban to wires and machines

Darkere:
They are using some weird stuff for their Itemregistration. 
Translationkey is "unnamed". That's why over resource location. 

Machines don't actually crash they just render nothing so that part is not strictly required.


raoulvdberge:
Can you convert this logic to another method and make some of the boolean logic clearer (adding if statements etc.)

Please also change the changelog entry as well :)

raoulvdberge:
I can't merge, your changelog is causing a conflict.

Darkere:
huh never used githubs conflict tool before I think that worked?

raoulvdberge:
thx

PR #2288, opened 2019-07-24
Fixed broken rendering of patterns

vildaberper:
Pattern rendering now goes through TileEntityItemStackRendererPattern as intended, but some GregTech stacks are still being handled differently. At least everything renders properly...


vildaberper:
Well, turns out some stuff are still "broken" but at least the client doesn't crash with mods using the old way of rendering (IItemRenderer) - patterns will simply render invisible.

raoulvdberge:
I can't say I'm happy with this approach. Is there not a more generic fix instead of literally checking if the mod is gregtech?

vildaberper:
Probably yes, but I'm not familiar with Forge at all and the rendering seems to be a mess overall. I'm pretty sure this is a step in the right direction however since it doesn't crash the client but simply renders patterns invisible. This is the proper way to forward the render but the issue is a reliable way to call a generic render on the resulting ItemStack.
In any case, it's better than the current solution which in my eyes makes this pull request valid.

raoulvdberge:
I'm not satisfied with this solution.

PR #2329, opened 2019-10-21
autocrafting performance improvements

Darkere:
~~still need to remove gradle...~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 1.6.16
 - Updated Russian translation (Bytegm)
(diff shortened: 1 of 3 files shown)


Darkere:
I arbitrarily picked 500 crafts for the feature to kick in. Not sure what value is appropriate here.

Also, there is another tiny change that makes it so that simulated insertions don't copy the stack.

Darkere:
...luckily commit messages from PR's don't appear in the end commit. That is just the changelog update.

raoulvdberge:
This won't be tweaked for 1.12 anymore

PR #2331, opened 2019-10-23
Fix quadratic behaviour when using ArrayLists

mcpower:
This should significantly speed up filtering in `GridViewBase` and crafting task updates.

Porting these fixes to the existing 1.14 branch should be easy.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/task/CraftingTask.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/autocrafting/task/CraftingTask.java
@@ -1,6 +1,7 @@
 package com.raoulvdberge.refinedstorage.apiimpl.autocrafting.task;
 
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.raoulvdberge.refinedstorage.RS;
 import com.raoulvdberge.refinedstorage.api.autocrafting.*;
 import com.raoulvdberge.refinedstorage.api.autocrafting.craftingmonitor.ICraftingMonitorElement;
@@ -590,13 +591,11 @@ public class CraftingTask implements ICraftingTask {
     }
 
     private void updateCrafting() {
-        Iterator<Crafting> it = crafting.iterator();
+        Set<Crafting> toRemove = Sets.newHashSet();
 
         Map<ICraftingPatternContainer, Integer> counter = Maps.newHashMap();
 
-        while (it.hasNext()) {
-            Crafting c = it.next();
-
+        for (Crafting c : this.crafting) {
             ICraftingPatternContainer container = c.getPattern().getContainer();
 
             int interval = container.getUpdateInterval();
@@ -649,7 +648,7 @@ public class CraftingTask implements ICraftingTask {
                         this.internalStorage.insert(byp, byp.getCount(), Action.PERFORM);
                     }
 
-                    it.remove();
+                    toRemove.add(c);
 
                     network.getCraftingManager().onTaskChanged();
 
@@ -657,20 +656,20 @@ public class CraftingTask implements ICraftingTask {
                 }
             }
         }
+
+        this.crafting.removeAll(toRemove);
     }
 
     private void updateProcessing() {
-        Iterator<Processing> it = processing.iterator();
+        Set<Processing> toRemove = Sets.newHashSet();
 
         Map<ICraftingPatternContainer, Integer> counter = Maps.newHashMap();
 
-        while (it.hasNext()) {
-            Processing p = it.next();
-
+        for (Processing p : this.processing) {
             ICraftingPatternContainer container = p.getPattern().getContainer();
 
             if (p.getState() == ProcessingState.PROCESSED) {
-                it.remove();
+                toRemove.add(p);
 
                 network.getCraftingManager().onTaskChanged();
 
@@ -780,6 +779,8 @@ public class CraftingTask implements ICraftingTask {
                 }
             }
         }
+
+        this.processing.removeAll(toRemove);
     }
 
     private static boolean insertIntoInventory(@Nullable IItemHandler dest, Deque<ItemStack> stacks, Action action) {
--- a/src/main/java/com/raoulvdberge/refinedstorage/gui/grid/view/GridViewBase.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/gui/grid/view/GridViewBase.java
@@ -46,19 +46,14 @@ public abstract class GridViewBase implements IGridView {
                 (grid.getTabSelected() >= 0 && grid.getTabSelected() < grid.getTabs().size()) ? grid.getTabs().get(grid.getTabSelected()).getFilters() : grid.getFilters()
             );
 
-            Iterator<IGridStack> it = stacks.iterator();
-
-            while (it.hasNext()) {
-                IGridStack stack = it.next();
-
+            stacks.removeIf(stack -> {
                 for (Predicate<IGridStack> filter : filters) {
                     if (!filter.test(stack)) {
-                        it.remove();
-
-                        break;
+                        return true;
                     }
                 }
-            }
+                return false;
+            });
 
             GridSorterDirection sortingDirection = grid.getSortingDirection() == IGrid.SORTING_DIRECTION_DESCENDING ? GridSorterDirection.DESCENDING : GridSorterDirection.ASCENDING;
 


Darkere (commented):
The Collections.removeIf function works exactly the same way as the code you deleted. Not sure what this is supposed to change.
~~Also not sure what the idea behind changing the ArrayList to a set is. Can you explain your thought process please?~~

mcpower:
> The Collections.removeIf function works exactly the same way as the code you deleted. Not sure what this is supposed to change.

Good point - the default [`Collection#removeIf`](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Collection.java#l409) does indeed work in the same way. However, [`ArrayList#removeIf`](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/ArrayList.java#l1378) optimises this out to be linear. I'm not sure which method `stacks.removeIf` calls, as it's defined as an `ArrayList` but its type is `List` - which defaults to the `Collection` implementation. I would assume it calls the `ArrayList` implementation - but I don't have much Java experience so I don't actually know. Should we change the type of `stacks` to be explicitly `ArrayList<IGridStack>` instead, just to be sure that the `ArrayList` optimisation?

> Also not sure what the idea behind changing the ArrayList to a set is. Can you explain your thought process please?

I didn't change any ArrayList to a set - do you mean the introduction of the `toRemove` set in `CraftingTask` methods? The code still iterates over the `crafting` and `processing` `ArrayList`s using a foreach loop, instead of explicitly using iterators.

mcpower:
I just checked using a debugger and yes, `stacks.removeIf` does indeed use the `ArrayList` implementation of `removeIf`.

raoulvdberge:
You're calling removeIf on an ArrayList. This doesn't cause any performance improvements, right?

mcpower:
Both removeIf and removeAll are O(n), assuming that the predicate/(collection's contains method) is O(1) (which it is in this case). The existing code exhibits O(n²) behaviour due to calling `it.remove()` up to n times - [which internally calls `ArrayList#remove`](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/ArrayList.java#l853) which is O(n) due to the reshuffling of the array items.

raoulvdberge:
Realistically, is the reshuffling of array items more expensive than iterating over all the elements and using a predicate?

raoulvdberge:
I think looking strictly at theoretical performance gains (with big o notation) is kind of dangerous. We probably need some hard performance numbers to see if this actually increases performance.

raoulvdberge:
I think looking strictly at theoretical performance gains (with big o notation) is kind of dangerous. We probably need some hard performance numbers to see if this actually increases performance.

raoulvdberge:
Thank you, I'll be looking forward to the performance benchmarks. I'm not so knowledgeable on the subject so I'd like to see some proof that this actually increases performance in practice. :)

mcpower:
I wrote a few benchmarks - source code [here](https://gist.github.com/mcpower/1c5196c9e596b9ee89387da17c9adb52):

- `iteratorRemove`: using `it.remove()`
- `listRemove` - using `ArrayList#remove`
- `removeIfRemove`: using `removeIf`
- `removeAllRemove`: using `HashSet<Object> toRemove` and `removeAll`
- `manualArrayListRemove`: using `ArrayList<Integer> toRemove` and doing some non-trivial work afterwards
- `manualBitsetRemove`: using `BitSet toRemove`
- `optimisedBitsetRemove`: using `BitSet toRemove` and using very similar code to OpenJDK's `removeIf`

Here are the results for a list of 1,000 items, removing 90% of them:

```
Benchmark                              Mode  Cnt   Score   Error  Units
ListRemoveBench.iteratorRemove         avgt   25  84.691 ± 0.464  us/op
ListRemoveBench.listRemove             avgt   25  84.655 ± 0.403  us/op
ListRemoveBench.manualArrayListRemove  avgt   25  11.737 ± 0.305  us/op
ListRemoveBench.manualBitsetRemove     avgt   25  10.630 ± 0.410  us/op
ListRemoveBench.optimisedBitsetRemove  avgt   25  11.116 ± 0.181  us/op
ListRemoveBench.removeAllRemove        avgt   25  29.668 ± 0.270  us/op
ListRemoveBench.removeIfRemove         avgt   25   2.946 ± 0.017  us/op
```

1,000 items, removing 10% of them:

```
Benchmark                              Mode  Cnt   Score   Error  Units
ListRemoveBench.iteratorRemove         avgt   25  13.887 ± 0.072  us/op
ListRemoveBench.listRemove             avgt   25  13.818 ± 0.042  us/op
ListRemoveBench.manualArrayListRemove  avgt   25   7.960 ± 0.057  us/op
ListRemoveBench.manualBitsetRemove     avgt   25   8.580 ± 0.368  us/op
ListRemoveBench.optimisedBitsetRemove  avgt   25   8.765 ± 0.131  us/op
ListRemoveBench.removeAllRemove        avgt   25  16.482 ± 0.048  us/op
ListRemoveBench.removeIfRemove         avgt   25   4.875 ± 0.097  us/op
```

1,000 items, removing 1% of them - this heavily favours fast code like `System.arrayCopy`:

```
Benchmark                              Mode  Cnt   Score   Error  Units
ListRemoveBench.iteratorRemove         avgt   25   5.096 ± 0.019  us/op
ListRemoveBench.listRemove             avgt   25   5.122 ± 0.015  us/op
ListRemoveBench.manualArrayListRemove  avgt   25   7.666 ± 0.054  us/op
ListRemoveBench.manualBitsetRemove     avgt   25   7.366 ± 0.105  us/op
ListRemoveBench.optimisedBitsetRemove  avgt   25   9.668 ± 0.176  us/op
ListRemoveBench.removeAllRemove        avgt   25  16.419 ± 0.228  us/op
ListRemoveBench.removeIfRemove         avgt   25   8.626 ± 0.039  us/op
```

10,000 items, removing 1% of them:
```
Benchmark                              Mode  Cnt    Score   Error  Units
ListRemoveBench.iteratorRemove         avgt   25  140.569 ± 1.267  us/op
ListRemoveBench.listRemove             avgt   25  141.387 ± 0.740  us/op
ListRemoveBench.manualArrayListRemove  avgt   25   71.467 ± 3.311  us/op
ListRemoveBench.manualBitsetRemove     avgt   25   80.488 ± 9.471  us/op
ListRemoveBench.optimisedBitsetRemove  avgt   25   80.537 ± 7.732  us/op
ListRemoveBench.removeAllRemove        avgt   25  145.571 ± 1.144  us/op
ListRemoveBench.removeIfRemove         avgt   25   88.148 ± 0.353  us/op
```

We can clearly see that `removeIf` is *very* fast, especially if removing lots of items - in fact, the more items that are removed, the faster it is.

`removeAll` with a `HashSet` performs pretty badly but still scales better than `it.remove()` and `ArrayList#remove` when removing lots of items. We could probably replace that with code similar to the `ArrayList`/`BitSet` implementations.

In the use case of filtering items in a grid view, it's likely that 90% or more items will be filtered - as you'd probably be searching for a specific item. `removeIf` wins out in this case.

I can't really tell about the other use case - crafting task updates - as I don't fully grok how many items are removed on average per update. It's likely that one of the `ArrayList`/`BitSet` implementations performs better, unless there's only a small percentage of removals per update. The whole implementation would get replaced in #2246 anyway.

Darkere:
#2246 will not get merged for 1.12 so this change would very much be relevant. Your benchmarks show similar results to my practice tests. 

Autocrafting where 90% of the craft gets removed after one iteration does not matter for performance. It would, after all, finish after 2 game ticks. Problematic is when a very large craftingrequest is made. They can take 10+ minutes to finish (in comparison a request that removes 1% of entries every iteration would finish in 5 seconds). 

Using a basic crafter the performance difference is negligible similar to your last benchmark. However, using a max sized multiblock crafter the results favor the iterator implementation with a performance difference between 30-40%. Again similar to your second and third benchmark.

mcpower:
So, as I understand it, we have two separate use cases:
- `GridViewBase#sort`, which often removes lots of items
- autocrafting, which often removes 1 or 0 items.

In both use cases, I don't believe there is any need for the order of the list to be maintained. For the first case, the list gets sorted immediately after it's filtered. For the second use case, I presume order does not matter ( I recently realised there's some additional ways of doing this filtering:
- `manualRemove` - removes items on-the-fly and doing a resize at the end
- `newArrayList` - creates a separate ArrayList for the filtered list and adds to that
- `unorderedSwap` - removes items on-the-fly **by changing the order of the list** and resizing
- `unorderedSwapBatchResize` - removes items on-the-fly **by changing the order of the list** and doing a resize at the end


1,000 items, removing a *single* item:
```
Benchmark                                 Mode  Cnt   Score   Error  Units
ListRemoveBench.iterator                  avgt   25   3.217 ± 0.041  us/op
ListRemoveBench.manualArrayList           avgt   25   7.073 ± 0.110  us/op
ListRemoveBench.manualBitset              avgt   25   6.892 ± 0.185  us/op
ListRemoveBench.manualRemove              avgt   25   2.743 ± 0.011  us/op
ListRemoveBench.newArrayList              avgt   25   7.934 ± 0.084  us/op
ListRemoveBench.optimisedBitset           avgt   25   7.560 ± 0.093  us/op
ListRemoveBench.removeAll                 avgt   25  12.358 ± 0.033  us/op
ListRemoveBench.removeIf                  avgt   25   9.414 ± 1.133  us/op
ListRemoveBench.unorderedSwap             avgt   25   4.403 ± 0.063  us/op
ListRemoveBench.unorderedSwapBatchResize  avgt   25   2.335 ± 0.015  us/op
```

10,000 items, removing a *single* item:
```
Benchmark                                 Mode  Cnt    Score   Error  Units
ListRemoveBench.iterator                  avgt   25   38.775 ± 0.238  us/op
ListRemoveBench.manualArrayList           avgt   25   69.007 ± 3.013  us/op
ListRemoveBench.manualBitset              avgt   25   57.795 ± 1.714  us/op
ListRemoveBench.manualRemove              avgt   25   31.363 ± 0.422  us/op
ListRemoveBench.newArrayList              avgt   25   69.817 ± 5.397  us/op
ListRemoveBench.optimisedBitset           avgt   25   66.823 ± 3.061  us/op
ListRemoveBench.removeAll                 avgt   25  111.443 ± 0.473  us/op
ListRemoveBench.removeIf                  avgt   25   87.189 ± 0.655  us/op
ListRemoveBench.unorderedSwap             avgt   25   46.579 ± 0.333  us/op
ListRemoveBench.unorderedSwapBatchResize  avgt   25   23.208 ± 0.087  us/op
```

Surprisingly, `unorderedSwapBatchResize` is significantly faster than `unorderedSwap` - even though we're removing a single item!

For fun, here's what happens when we invert the predicate:

1,000 items, removing all but one item:
```
Benchmark                                 Mode  Cnt   Score   Error  Units
ListRemoveBench.iterator                  avgt   25  97.795 ± 0.479  us/op
ListRemoveBench.manualArrayList           avgt   25  11.245 ± 0.181  us/op
ListRemoveBench.manualBitset              avgt   25  11.427 ± 0.173  us/op
ListRemoveBench.manualRemove              avgt   25   2.660 ± 0.018  us/op
ListRemoveBench.newArrayList              avgt   25   2.585 ± 0.028  us/op
ListRemoveBench.optimisedBitset           avgt   25   8.868 ± 0.342  us/op
ListRemoveBench.removeAll                 avgt   25  27.018 ± 0.407  us/op
ListRemoveBench.removeIf                  avgt   25   2.803 ± 0.050  us/op
ListRemoveBench.unorderedSwap             avgt   25   4.995 ± 0.012  us/op
ListRemoveBench.unorderedSwapBatchResize  avgt   25   3.360 ± 0.016  us/op
```

```
Benchmark                                 Mode  Cnt     Score    Error  Units
ListRemoveBench.iterator                  avgt   25  8217.226 ± 27.427  us/op
ListRemoveBench.manualArrayList           avgt   25   109.941 ±  1.517  us/op
ListRemoveBench.manualBitset              avgt   25    87.351 ±  8.462  us/op
ListRemoveBench.manualRemove              avgt   25    24.878 ±  0.109  us/op
ListRemoveBench.newArrayList              avgt   25    26.306 ±  0.524  us/op
ListRemoveBench.optimisedBitset           avgt   25    79.450 ±  7.579  us/op
ListRemoveBench.removeAll                 avgt   25   230.224 ± 10.950  us/op
ListRemoveBench.removeIf                  avgt   25    27.757 ±  0.637  us/op
ListRemoveBench.unorderedSwap             avgt   25    49.703 ±  0.371  us/op
ListRemoveBench.unorderedSwapBatchResize  avgt   25    32.422 ±  0.191  us/op
```

raoulvdberge:
Thank you, I've implemented the grid change in the 1.15 branch. The autocrafting code remains unchanged for clarity.

PR #2347, opened 2019-11-08
1.14 Autocrafting Rework

Darkere:
Ports #2246 to 1.14

<details>
  <summary>Summary of Changes</summary><p>

- Sorting of both the Crafting Monitor Display and the Crafting preview

- implemented insert order for Patterns #1937

Due to this change, these to Recipes are no longer the same: 
image

In fact, the right recipe will not work at all as the machine itself only has 2 slots and there will be no attempt at inserting a combined set. 

improved Performance for: 
- Crafting Calculation
- Craft updating
- NBT read and write
- Item importing while crafting tasks are running

This contains API changes.

</p>
</details>

EDIT: Additional Change I just realized. Removing patterns from a running autocraft will now stop that craft from continuing.


raoulvdberge:
As discussed on Discord I'll be closing this PR as you are going to make a new PR with all the changes nicely condensed per-commit, instead of one big blob of code changes.

PR #2352, opened 2019-11-14
Autocrafting Rework 

Darkere:
As discussed, the rework split into parts.
For now, I've planned 6 parts/commits and 3 additional minor PR's. 
I've not found a way to submit the PR's before this one is merged, do tell me if there is a way to do that. 

Part 1:
This one doesn't actually change much. Only cleans up the code for the update methods.
I've moved everything that does item/fluid extraction/insertion into their own methods. 
This forces a couple of changes to where errors are detected but should overall clear up what is happening a bit.


Darkere:
Part 2: More prepwork
This combines the common parts of Crafting and Processing into a superclass. While they currently only have a few things in common, that is going to change in further commits. 
Some things like `counter` and `toRemove` are a bit clunky in this version but I just added them to check if I broke anything.

Darkere:
Part 3: Calculation
Now we get to the parts that actually make a difference. Fundamentally what changes is that instead of calculating everything for every step of the crafting, it now calculates things for every pattern once and adds multipliers to get match up how many times everything is needed. 

This changes quite a few things: 
- it's necessary to calculate how many times a pattern needs to be crafted every step, for this, I've repurposed the `getQuantityPerCraft` method
- loops around `calculateInternal` have become completely unnecessary 
- it's possible to get Item's in part from `fromSelf`, `fromNetwork` and from crafting so that needed to be adjusted
- PatternChains no longer work at all. The ability for multiple crafters to do the same thing is coming back later

Darkere:
Part 4: More speed up for calculation
Now that the calculation supports multiple items there is another optimization that can be done. Since we only really need the actual recipe (I renamed `took` to recipe) for figuring out the output of the crafting, we can sum up all the ingredients that are needed to reduce the number of times we have to go through calculation. 
This makes any recipes that have a lot of the same ingredients much faster. Best case would be things like octuple compressed cobblestone going from 9^8 calculation calls to just 9.

Darkere:
Part 5: fixing Crafting/Processing
Crafts now only get updated once every update and contain all the items so It needs to be adjusted to work with that. 
For Crafting that means the counter is gone and instead it loops once for each successful crafting update. Then it's just counting down the quantity. 

Processing is a bit more complex. Here we need to keep track of :
- how many times items still need to get inserted (quantity)
- how many times items have been inserted (total - quantity)
- how many complete sets have finished processing (finished)

Finished is calculated by figuring out the number of patterns that have already received the items necessary to complete. This forces changing how tracked items are accounted for.
Everything else is just multiplying item counts in the right places. 

The items that are used aren't properly calculated yet so that breaks still.

Darkere:
Part 6: Oredict/Tag System
This is the tricky part. All calculated items need to be added as potential items that may get used. However, the system needs to know how many of which type of item it needs to insert every iteration too. 

The cleanest solution I have found stores all items that may be used for one ingredient inside a stacklist. Then when extracting it grabs the items for each ingredient from the associated stacklist.

Fluids are easier they work almost the same way as before. I just had to add them in a bit earlier. 

I've also moved where crafts are created to the beginning of calculation. This way it's more obvious which parts of the code are for calculation and which for after. It also sorts the crafts in a logical order which will be used later by the crafting monitor.

Darkere:
**Part 7: Multiple Container and Hashmaps** 

Looping over each container that contains the pattern reintroduces the speedup when multiple containers have the same pattern. For this, it maintains a list for each pattern that has all containers listed. As a consequence autocrafting now also responds to adding/removing patterns while a task is running.

There is a rather minor issue that necessitates the introduction of the Linkedhashmap for crafts. When the same craft is requested multiple times in different parts of the task then the crafting speed would multiply. Solved by only adding quantity and ingredients when the same pattern is found. 

(For the PatternChain you deliberately introduced a new class `Key` to avoid overwriting `equals` and `hashCode` for `CraftingPattern`. I'm not seeing the reason for that but it could be reintroduced rather easily.)

There are still a couple of minor fixes left to do. And some cleanup after that.

Darkere:
**Part 8: Small fixes**
- Fix totalsteps. 
- change ItemsToUse to a Map to avoid ArrayList nonsense
- add ProcessingState support for multiple containers. The state now only changes if none of the container were able to run
- fix Crafting Monitor updates happening too early for tracked items. 
- Crafting Monitor now shows the needed items instead of multiplied outputs. It makes it more clear what the processing is waiting for.

Darkere:
And with that everything is up and running. Gonna do more extensive testing so there may be more but for now, this is everything.

There are still parts missing that are going into separate PR's:
- Sorting of the Crafting Monitor
- Sorting of the Crafting Preview
- keep pattern order for processings
- improve speed of tracked insertions

raoulvdberge:
OK! Will check this out ASAP. Please wait with the next PRs until this one is merged.

raoulvdberge:
closing in favor of the 1.15 pr

PR #2363, opened 2019-12-02
Fix extractFluid voiding excess in OC driver

BlueAgent:
## Description
Using the OC driver, if you extract more fluid than can fit in a tank it would void the excess.

This PR makes it so it only extracts from the network as much as can be inserted into the tank.

In the edge case that a tank varies its input limit based on how much fluid is being inserted, the excess fluid will attempted to be inserted back into the network.

## Mod Versions (used for testing)
minecraft-1.12.2
forge-14.23.5.2847
refinedstorage-1.6.15
OpenComputers-MC1.12.2-1.7.5.192
extrautils2-1.12-1.9.9

## Demonstration Videos
[Before Fix](https://www.youtube.com/watch?v=_wUAqh8vWo0)
[After Fix](https://www.youtube.com/watch?v=iNNFwOVLhzk)


raoulvdberge:

PR #2379, opened 2019-12-18
Potential fix NPE in ItemGridView::postChange occurring on client.

noobanidus:
This appears to happen when the contents of the grid previously
contained items that were also ingredients for an autocrafting recipe,
and the autocrafting recipe has been created.

Based on the null check in the block of code after the one edited, it
seems to be a thing that the map can potentially return a null result;
however, the "update other id" block does not test for null, simply
calling `updateOtherId`.

This commit adds a test for a potential null result from the map.

While this might just be addressing the side effect of a larger issue
associated with auto-crafting (i.e., the expectation that the IGridStack
wouldn't be a null result), it hopefully reduces the number of annyoing
client-side crashes.

-> After some extensive testing in a situation where I was getting multiple crashes per gaming session, using this modification seems to have reduced them to nothing.

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/screen/grid/view/ItemGridView.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/screen/grid/view/ItemGridView.java
@@ -32,7 +32,10 @@ public class ItemGridView extends BaseGridView {
         // With that new id, the reference for the crafting stack would be outdated.
         if (!stack.isCraftable() &&
             stack.getOtherId() != null) {
-            map.get(stack.getOtherId()).updateOtherId(stack.getId());
+            IGridStack result = map.get(stack.getOtherId());
+            if (result != null) {
+                result.updateOtherId(stack.getId());
+            }
         }
 
         ItemGridStack existing = (ItemGridStack) map.get(stack.getId());


Darkere:
If you look at the issue in a debugger you can see that the actual issue happens when the map is completely empty. 

Your solution will also catch this situation. I've just put a catch in for empty and similarly not had any issues since. 

We need to investigate if it is at all related to the craftable condition and whether or not the message can be discarded entirely or needs to be added to the map immediately or later.

noobanidus:
I didn't have it reproducible (or rather, I didn't go to the lengths needed to be able to reproduce it) in such an environment, so I wasn't sure whether it was the map being empty or whether there were situations where the map had information in it, but not specifically this information; judging by the future null checks, it seemed sensible to also check for null here.

Definitely the issue only ever seems to occur for me (and the other person using my refined storage network) when opening a crafting grid immediately after having performed an autocrafting procedure on another terminal.

raoulvdberge:
This is not the right fix. This NPE check is masking an actual issue. 

While I can't reproduce this, I'm assuming this crash is caused by this:

1) Open a Grid
2) When the Grid is open, a Crafter is causing an invalidation of the craftables list (for some reason??? Could be a) a new pattern was inserted b) crafter got disconnected and reconnected (energy problems?) or c) the direction of the crafter changed)
3) This invalidation causes a rebuild of the craftables list, giving new IDs
4) As soon as a Grid sends a delta packet with an item change, this new ID is sent along as well
5) The client never knew of any ID changes to the craftables list, they only get that when the grid is initially opened.

So while I'm speculating, I think that's the issue. I'll implement a fix for that scenario next release and if it's not helping, we can resort to null checks.

raoulvdberge:
I don't believe this fixes the actual issue. This might mask an issue. I'll close this PR for now while we investigate the actual cause of this. If we can't find a cause, we can resort to null checks.

raoulvdberge:
Update: we found the cause. Turns out this PR could've still worked.

The problem was that the Grid is sending delta packets when the client is watching the screen.

It stops sending those as soon as the client closes the screen. But as this is serverside, there is some latency till the server knows the client has closen the screen.

This means that the server might still be sending delta packets when the client has closen the screen.

Due to our screen `executeLater` system, those "rogue" old packets would still be handled as soon as we open the Grid screen again moments later.

Because `postChange` (delta packet) would be called before `setStacks` (initial update packet), it would crash with that NPE because the map is still empty.

PR #2387, opened 2019-12-27
Fixes Startup crash loop edge case

### LINKED ISSUE #2381: Edge case not accounted for that will cause a server to crash on startup
with the current code on update apiimpl/network/Network.java line 173
```
world.setBlockState(pos, world.getBlockState(pos).with(ControllerBlock.ENERGY_TYPE, energyType));
```

is called, in the rare case where the block you are setting the state of is actually no longer a controller this causes a crash because of a java illegal argument exception.

although for the most part this shouldn't happen I have seen it, specifically in the case where the mod has been removed and re added, the node and network data files still exist and point to a location in world being a controller but said location is now minecraft:air.

#### Issue description: 
edge case causing server crash

#### What happens: 
attempt to assign an enum property on a block that doesn't contain that property

check that the block is in fact what it is meant to be

#### Steps to reproduce:
1. unknown
...

- Minecraft: 1.14.4
- Forge: 28.1.107
- Refined Storage:

Does this issue occur on a server? [yes/no] yes

```
java.lang.IllegalArgumentException: Cannot set property EnumProperty{name=energy_type, clazz=class com.raoulvdberge.refinedstorage.block.ControllerBlock$EnergyType, values=[off, nearly_off, nearly_on, on]} as it does not exist in Block{minecraft:air}
	at net.minecraft.state.StateHolder.func_206870_a(SourceFile:106) ~[?:?] {re:classloading,pl:accesstransformer:B}
	at com.raoulvdberge.refinedstorage.apiimpl.network.Network.update(Network.java:173) ~[?:1.7.1] {re:classloading}
```

full link
https://pastebin.com/s4ZSfaJe

LezChap:
Fixes #2381

### DIFF
--- a/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/Network.java
+++ b/src/main/java/com/raoulvdberge/refinedstorage/apiimpl/network/Network.java
@@ -28,6 +28,7 @@ import com.raoulvdberge.refinedstorage.apiimpl.storage.cache.ItemStorageCache;
 import com.raoulvdberge.refinedstorage.apiimpl.storage.tracker.FluidStorageTracker;
 import com.raoulvdberge.refinedstorage.apiimpl.storage.tracker.ItemStorageTracker;
 import com.raoulvdberge.refinedstorage.block.ControllerBlock;
+import com.raoulvdberge.refinedstorage.config.ServerConfig;
 import com.raoulvdberge.refinedstorage.energy.BaseEnergyStorage;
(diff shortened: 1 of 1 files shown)


raoulvdberge:
Thx for the PR, I implemented your changes in 3915201cf284b6745aa90ef890fdd770339f1f2b. Difference between that commit and this PR is that there I'm doing the `getBlockState` call once.

LezChap:
Your code is cleaner and smarter.  I was just trying to get something out that worked and stopped the crashloop. :)

PR #2397, opened 2020-01-11
b485c74503b84cb7b7f7f105ecbce566be9f520e backport

Dabombber:
I'm not sure if any of the later fixes should be included or if they're regressions from the 1.14 port, but I've found this bug particularly irritating.

Tested this and it seems to work, but there's a lot of internal changes so I might have missed something.


raoulvdberge:

PR #2405, opened 2020-01-25
small recipe fix

Darkere:
Recipes made without exact mode on may fall back to the exact ingredients if the recipe does not provide alternatives.

Fixes
image


raoulvdberge:
Can you remove the language key as well and revert the formatting changes?

Darkere:
hmm those must have sneaked in when I auto formatted :/ need to fix that. your version is better

raoulvdberge:
Nice thx!

PR #2404, opened 2020-01-22
1.15autocrafting

Darkere:
Ported #2352 to 1.15.

refer to #2352 for comments

PR #2406, opened 2020-01-27
Autocrafting engine v6

Darkere:
Change list:
- Added a config option for switching between autocrafting engines
- Added #2404 under the v6 package
- Added sorted CraftingPreview
    - done by rearranging the order things are processed and by moving `toCraft/toCraftFluids` to an Arraylist so they are sorted and reversible
- Added sorted CraftingMonitor
    - done by splitting the element list into 3 separate lists and merging them when appropriate
    - there are some additional minor changes that aim to reduce flickering inside the crafting monitor

Craftingmonitor is still not doing things as I want it to... but I've resigned myself to the fact that it's not possible without rewriting how errors work in the crafting monitor. I may do that at some point in the future. It's good enough for now.


Darkere:
Been testing this on a public server for the last month.

Fixed all the issues that have been found. And managed to finally make the crafting monitor do what I want it to, without affecting v5.

raoulvdberge:
Let's do this!

Issue #2014, opened 2018-09-29
Inventory Tweaks middle-click sorts like the player's inventory is one big container (in the Grid)


liquid600pgm:
### Issue description:
When middle-clicking on my inventory to sort it (in the Grid), the sorting treats the hotbar and "backpack" portions as one.

### What happens:
![What happens](https://i.imgur.com/deuJVYl.png)

### What you expected to happen:
![What I expect](https://i.imgur.com/VxZD6bg.png)

#### Proper sorting
1. Open your inventory
2. Middle-click on your inventory to sort

#### Broken sorting
1. Open a (Crafting) Grid
2. Middle-click on your inventory to sort

### Versions:

- Minecraft: 1.12.2
- Forge: 14.23.4.2759
- Refined Storage: 1.6.5
- Inventory Tweaks: 1.63

Does this issue occur on a server? - No

DarkHelmet433:
Sadly this is a duplicate of #1542 and #1933 - I think Raoul is expecting the Inventory Tweaks author to look into this, but it's not clear that anything is happening.

Personally, I delete Inventory Tweaks and replace it with Inventory Sorter by cpw.  That mod actually works.

raoulvdberge:
Yes, I've had enough reports about this. Now it's time for  the InventoryTweaks author to step in and guide me further, as I don't believe this is an RS problem.

PR #2447, opened 2020-03-05
[1.12] Fix Inventory Tweaks woes

Landmaster:
There has been a long-standing bug with Inventory Tweaks (#2014, Inventory-Tweaks/inventory-tweaks#500) where sorting in a Refined Storage grid would not respect locked slots, and this PR fixes that. The problem was that you added the hotbar slot first, followed by the other slots, whereas it really should be the other way around.


Landmaster:
@raoulvdberge Any decision on this PR yet?

raoulvdberge:

Issue #2394, opened 2020-01-07
All nodes reverting when leaving the nether without causing a save


Darkere:
#### Issue description:
All nodes revert to default or a previous state when leaving through a nether portal. 
#### What happens:
Disk drives empty, importer settings gone. Disk data itself seems fine. 

#### Steps to reproduce:
1. Create RS system in the Nether
2. Leave through portal before a save happens
3. come back couple seconds later. 
...

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.14.4
- Forge: 28.1.109
- Refined Storage: 1.7.3

zedekblue:
Just had this issue occur on my world. 64k disk from the drive I had is gone.
Left the nether via an rftools teleporter.

Darkere:
Fixed in 1.16

PR #2552, opened 2020-06-05
implement safe writing for the disk file fixes #1846

### LINKED ISSUE #1846: (BUG) All Refined Storage Drive Bays empty after server restart
#### Issue description:
We perform scheduled restarts of our servers twice a day, just to keep performance 'fresh'. After this morning's scheduled restart, the server came back online with all RS Drive Bay blocks being empty of drives.

Not this.

#### Steps to reproduce:
1. Literally no clue what caused it.

- Minecraft: 1.12.2
- Forge: 2705
- Refined Storage: 1.5.34
- Refined Storage Addons: 0.3
- Reborn Storage: 1.12.2-3.1.0.50

Does this issue occur on a server? YES

No crash log, but we are trying to pull the server logs from this time period. It's going to be a metric butt-ton of data to go thru tho. Want 'em ?

Darkere:
Seems to be working just fine. Double-checked in NBT explorer that the data is formatted the exact same way. 

This solution first saves to a temporary file, then deletes the original file and renames the temporary file to the original filename. 

Still, have a couple concerns: 
- The DataFixer stuff I copied from WorldDimensionSaveManager looks like it might need to be updated manually? Not sure if that even matters if we aren't using datafixers in the first place. 

- Same as the Normal WorldSavedData I used the compressed version of writing NBT. Considering how common the crash was while it was saving, it might be a good idea to speed it up by not compressing the data?


raoulvdberge:
This PR is very invasive. So I see you have a solution, but what is the problem here? What causes it to happen and why does this solution fix said problem?

Darkere:
The Problem:
When WorldSavedData gets saved the current save file gets directly overridden. This is probably done to increase performance. 
This makes it susceptible to corruption if the game ever hard crashes while the file gets written. Vanilla can work with this by keeping the files very small and not keeping any critical data (anything important gets saved in level.dat instead).

This PR changes the way disk files get saved. They first get saved to a temporary file and only after this the original file gets deleted. This avoids the issue entirely. 
If the game crashes while writing the temp file, we can just go on with the previous save.

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -7,32 +7,57 @@
...
     private static final String NBT_DISK_ID = "Id";
     private static final String NBT_DISK_TYPE = "Type";
     private static final String NBT_DISK_DATA = "Data";
+    private static final String NBT_MINECRAFT_DATA = "data"; //native minecraft nbt string
+    private static final String NBT_MINECRAFT_DATA_VERSION = "DataVersion"; //native minecraft nbt string
 
     private final Map<UUID, IStorageDisk> disks = new HashMap<>();
     private final ServerWorld world;
+    private boolean dirty;
+    private static StorageDiskManager manager;
```
raoulvdberge, line 40: private static INSTANCE

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -7,32 +7,57 @@
...
-public class StorageDiskManager extends WorldSavedData implements IStorageDiskManager {
-    public static final String NAME = "refinedstorage_disks";
+public class StorageDiskManager implements IStorageDiskManager {
+    public static final String NAME = "refinedstorage_disks.dat";
 
     private static final String NBT_DISKS = "Disks";
     private static final String NBT_DISK_ID = "Id";
     private static final String NBT_DISK_TYPE = "Type";
     private static final String NBT_DISK_DATA = "Data";
+    private static final String NBT_MINECRAFT_DATA = "data"; //native minecraft nbt string
```
raoulvdberge, line 34: Might want to annotate this with `@todo remove` since I assume these tags are for compat? We probably want to remove these for 1.16+.

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -90,31 +115,68 @@ public void remove(UUID id) {
...
     @Override
-    public void read(CompoundNBT tag) {
-        if (tag.contains(NBT_DISKS)) {
-            ListNBT disksTag = tag.getList(NBT_DISKS, Constants.NBT.TAG_COMPOUND);
-
-            for (int i = 0; i < disksTag.size(); ++i) {
-                CompoundNBT diskTag = disksTag.getCompound(i);
+    public void read() {
+        CompoundNBT tag = readTagFromFile();
+        if (tag.contains(NBT_MINECRAFT_DATA)) {
```
raoulvdberge, line 124: Might want to do a warn if the file is corrupt. (e.g. doesn't contain necessary tags)

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -90,31 +115,68 @@ public void remove(UUID id) {
...
-                IStorageDiskFactory factory = API.instance().getStorageDiskRegistry().get(new ResourceLocation(type));
-                if (factory != null) {
-                    disks.put(id, factory.createFromNbt(world, data));
+        CompoundNBT nbt = null;
+        try {
+            if (diskFile.exists()) {
+                nbt = CompressedStreamTools.readCompressed(new FileInputStream(diskFile));
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
raoulvdberge, line 153: Use a logger here

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -90,31 +115,68 @@ public void remove(UUID id) {
...
                 }
+            } catch (IOException e) {
+                e.printStackTrace();
+                LOGGER.warn("Unable to read disk file and backup file. Continuing without disks loaded");
             }
         }
+        if (nbt == null) {// either no files have been generated yet or both files are broken
+            return new CompoundNBT();
+        }
+        //copied from DimensionSavedDataManager
```
raoulvdberge, line 170: Mark as @Volatile

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
```diff
@@ -126,9 +188,29 @@ public CompoundNBT write(CompoundNBT tag) {
...
+        CompoundNBT diskData = new CompoundNBT();
+        diskData.put(NBT_DISKS, disks);
+        CompoundNBT compoundNBT = new CompoundNBT();
+        compoundNBT.put(NBT_MINECRAFT_DATA, diskData);
+        compoundNBT.putInt(NBT_MINECRAFT_DATA_VERSION, SharedConstants.getVersion().getWorldVersion());
+
+        try {
+            writeTagToFile(compoundNBT);
+        } catch (IOException e) {
+            e.printStackTrace();
```
raoulvdberge, line 200: use a logger here

raoulvdberge:
Good idea. I added some comments. It might be a good idea to extract the NBT reading/writing logic so we can reuse this for network nodes and networks?

src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/StorageDiskManager.java
Darkere: Any corruption should be caught before this by an IO exeption in `readTagFromFile`.  This would also get called whenever it is read before it's generated.

Darkere:
Yep. I feared you would want to do that. Don't have enough time for that today. 

Any idea where to put the class for the safe writing logic? (and what to call it?)

raoulvdberge:
Probably in util. SaveDataManager or something?

Darkere:
I remembered why I put off making this work for all files. It's quite a bit of work.

As this now saves the data based on the dimension type instead of individual world objects. It also (at least it seems to) fix #2394

I tested it and it seems to be working but for this one you want to make sure yourself.

Darkere:
Closing in favor of #2607

PR #2557, opened 2020-06-06
reserve items used by CraftingTask to prevent using them more than on…

### LINKED ISSUE #2155: Exporter crafting task gets stuck
#### Issue description:

I have \*two* exporters set to export Diamond Chisels, with the Crafting Upgrade. After enough time, the crafting task for a Diamond Chisel will jam, with either a diamond, stick or oak wood planks missing in the Crafting Monitor. Manually crafting the missing item unjams the crafting task.

#### What happens:
Exporter crafting gets stuck

Exporter crafting should not get stuck

#### Steps to reproduce:
1. Install Chisel and Extra Utilities 2 (useful for demonstrating the bug)
2. Set up an RS network with lots of oak wood, diamond blocks and nothing else
3. Set up the following crafting recipes:
* 1 Oak Wood -> 4 Oak Wood Planks
* 2 Oak Wood Planks -> 4 Sticks
* 1 Diamond Block -> 9 Diamonds
* 1 Stick and 1 Diamond -> 1 Diamond chisel
4. Set up two Exporters for Diamond chisels, each with a Crafting Upgrade
5. Point the Exporters at a Trash Can or something to absorb the items
6. Wait long enough

- Minecraft: 1.12.2
- Forge: 1.12.2-14.23.5.2768
- Refined Storage: 1.6.11

Darkere:
…ce fixes #2155

This went too well. Just wrote it and it's working, which seems too convenient. 

I'm also not sure why exporter exporting the ingredients are not able to "steal" the ingredients from a crafting task. Some testing shows that this just doesn't happen. 

I assume it's due to the update order. Great as I would have had to rewrite the entire thing to make it work otherwise.


raoulvdberge:
It's possible that tasks steal from each other since autocrafting runs independantly from network node ticking (in different tick handlers). Also, when a device like an exporter schedules a task, it is added to a "to-add" list. The task isn't ticked right away, it's only ticked when the "to-add" list is cleared and when it's added to the current task lists. So, if devices schedule their tasks before any proceding task *actually* runs, it's entirely possible that the calculation makes wrong assumptions.

We could fix the root cause. One idea I have is to only allow 1 scheduled task in this "to-add" list at all times, but I'm not sure whether we want that. Another idea is to extract the initial items straight away, before the task is moved from the "to-add" list to the actual list. I don't know how good these ideas are, your fix might be better.

What do you think?

Darkere:
> One idea I have is to only allow 1 scheduled task in this "to-add" list at all times,

I figured this would potentially slow down Crafting in a way that is not possible to communicate to the player. 

> extract the initial items straight away

I was not certain why initial extraction happens where it currently does. The code seems very prepared for the fact that it might not be able to extract all the items it needs at once (and will continuously retry extracting the remaining items). 
Any items that could not be extracted on the first try could then be reused by the next task. 

That said this solution would certainly be much easier to implement.

raoulvdberge:
> That said this solution would certainly be much easier to implement.

With "this solution" do you mean, this PR, or the "extract when scheduled" fix?

Darkere:
I mean the "extract on scheduled" or rather extract directly after calculation. 

From memory this should be as simple as adding `extractInitial` if nothing is missing to the end of `CraftingTask.calculate`

raoulvdberge:
I think we should do that instead. 

Additionally, it might be helpful to get this to reproduce. So we know for sure that fixes it.

raoulvdberge:
Perhaps force task creation on same tick in the code?

Darkere:
Thoughts on renaming `CraftingManager#add` to `CraftingManager#start` and adding a start method into the crafting task?

Could also call this https://github.com/refinedmods/refinedstorage/blob/2933dee84906f2f9860f2409d380983461403c3b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java#L268
and this 
https://github.com/refinedmods/refinedstorage/blob/2933dee84906f2f9860f2409d380983461403c3b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java#L934
there.


This is already working just fine. Setup is just 14 exporter crafting the same thing at full speed. Making sure they all start at the same time by inserting the pattern after setting everything up.

raoulvdberge:
Yeah, good idea. I would also rename `CraftingTask#extractInitial` to `CraftingTask#start` then, and then add it to the `ICraftingTask` interface so it's part of API.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/Craft.java
```diff
@@ -138,4 +138,8 @@ CompoundNBT writeToNbt() {
         return tag;
     }
 
+    void finishCalculation() {
```
raoulvdberge, line 141: ?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java
```diff
@@ -596,6 +586,19 @@ private ICraftingTaskError calculateInternal(
         return null;
     }
 
+    @Override
+    public void start() {
+        crafts.values().forEach(craft -> {
```
raoulvdberge, line 591: Isn't the missing.isEmpty step needed?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java
```diff
@@ -596,6 +586,19 @@ private ICraftingTaskError calculateInternal(
...
+            craft.finishCalculation();
+        });
+
+        executionStarted = System.currentTimeMillis();
+
+        extractInitial();
+    }
+
+
     public void extractInitial() {
```
raoulvdberge, line 602: This can be private again I think?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/Processing.java
```diff
@@ -48,6 +48,7 @@
         this.itemsToDisplay = CraftingTask.readItemStackList(tag.getList(NBT_ITEMS_TO_DISPLAY, Constants.NBT.TAG_COMPOUND));
     }
 
+    @Override
```
raoulvdberge, line 51: ?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/Craft.java
Darkere: just cleans up the call structure a bit.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java
Darkere: oh ya. forgot

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingTask.java
Darkere: hmm it's not needed but probably a good idea to put one there.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/Processing.java
Darkere: part of the cleanup for the `finishCalculation` call. Now a bit prettier :)

raoulvdberge: but why is it suddenly `@Override`? I don't see this method in `Craft`?

PR #2561, opened 2020-06-07
minor changes to shift crafting behavior

Darkere:
This makes it so that if you have stacks with stack size higher than 1 in the crafting grid, shift crafting will now iterate so much to consume that amount and then stop. 

This allows for conveniently crafting amounts that aren't `maxStackSize` or require lots of clicking (particulary problematic on poor connections). 

Also fixes a small annoyance where shift crafting would produce more than one stack if it doesn't exactly hit `maxStackSize`. 

Demonstration: https://gfycat.com/FlakyDismalFlee


raoulvdberge:
I don't really understand this change, nor do I understand your explanation.

> This makes it so that if you have stacks with stack size higher than 1 in the crafting grid, shift crafting will now iterate so much to consume that amount and then stop.

So according to this explanation it would consume ALL the items? Or leave exactly 1 set of items in the slots?

> This allows for conveniently crafting amounts that aren't `maxStackSize` or require lots of clicking (particulary problematic on poor connections).

Shouldn't shift crafting always create items with size = maxStackSize? How does this change fix lots of clicking?

> Demonstration: https://gfycat.com/FlakyDismalFlee

So it leaves 1 set in the slots, what is the use of that? MC shift clicking consumes all items.

raoulvdberge:
I'm all for the second change, however I'm not convinced about the first change. I like the consistency with MC. I prefer clicking 32 times rather than filling up 1 random inventory slot so it can make 32.

Darkere:
I guess it is a somewhat obscure change. And you can also get around it by using autocrafting. fixed

Issue #2540, opened 2020-06-01
Item duplication with shulker boxes


derekrfelson:
#### Issue description:

A constructor set to non-exact mode will duplicate shulker boxes and their contents.

#### What happens:

The system sees an empty shulker box in storage and will try to place it, but what it actually places is the version of the shulker box (and its contents) that you configured the constructor with. The empty shulker box is not removed from storage, allowing you do dupe items and shulker boxes indefinitely.

#### What you expected to happen:

The constructor should place the empty shulker box, and the shulker box should be removed from storage.

#### Steps to reproduce:

1. Have a shulker box with the items you want to dupe, e.g. 10 diamonds.
2. Set up a constructor with exact mode off. Configure it to place the shulker with the items.
3. Empty the shulker box and place it in your storage system.
4. The constructor places a shulker box with the duped items while leaving the empty shulker box in storage. You just got a free shulker box and free items.
5. If you break the duped shulker box, the constructor will replace it and all the items, allowing you dupe as much as you want.

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.15.2 Java
- Forge:  forge-31.2.0
- Refined Storage: refinedstorage-1.8.3.jar

Does this issue occur on a server? yes

#### If a (crash)log is relevant for this issue, link it here:

N/A

PR #2584, opened 2020-06-21
fix ConstructorNetworkNode not extracting on placement

Darkere:
Didn't think onPlaceItemIntoWorld would modify the stack.


raoulvdberge:
Doesn't this reintroduce #2540?

Darkere:
No. The placed ItemStack still comes from the result of the Simulated extraction. The Performing extraction will then access the same stack the simulation did.

PR #2618, opened 2020-07-15
prevent unloaded networks from ticking

Darkere:
no more free chunkloading

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/api/network/INetwork.java
+++ b/src/main/java/com/refinedmods/refinedstorage/api/network/INetwork.java
@@ -40,6 +40,11 @@ public interface INetwork {
     NetworkType getType();
 
     /**
+     * updates the networks load status
+     */
+    void checkLoaded();
+
+    /**
      * @return true if this network is able to run (usually corresponds to the redstone configuration), false otherwise
      */
     boolean canRun();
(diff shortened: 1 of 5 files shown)


raoulvdberge:
What about network nodes? Those tick independantly of the network. I think you can still get free chunkloading that way.

Darkere:
Hmm, that makes that significantly more complicated. The check is not that cheap. So we can't check that before every node update. 

Caching it ends up with some nodes updating and some not. 

What do you think of changing it? Making it so when the network updates it updates all the nodes?

Darkere:
Actually. There is a somewhat simpler solution.

raoulvdberge:
Nodes need to be able to be updated without being in a graph.

Darkere:
oh. Under which circumstances? I guess if the network is not loaded but a node is then it needs to turn off...

Alright new idea.

Darkere:
This might work instead?

raoulvdberge:
That's not very clean code, especially the extra interface method.

Why don't you just check if the block is loaded before network & network node update? I see no need to "physically" turn the network off when the block is unloaded.

Darkere:
> I see no need to "physically" turn the network off when the block is unloaded.

You are assuming that the entire network is unloaded. If someone extends cables far enough that the controller unloads they can create nodes that look active but aren't. With extra confusion as they become active/inactive as they move around. 


> NetworkNodeManager has no business updating networks

??? I did not change anything about the way networks update. Just added a check before the update happens that looks if the network is loaded. 

The reason I made it so is because I thought the check is quite expensive. 
It loops through all loaded chunks... But only if you check on the client. 
Serverside it just gets the chunk from a map.
So yeah can totally check that for every node no problem. Will do in a sec.

PR #2623, opened 2020-07-17
Rework crafting storage and implement pattern reconstruction

Darkere:
I've been sitting on this for too long. Seems like a good time to do it. 

This addresses a bug where if a recipe has the option between 2 byproducts the crafting may give back the wrong items. 

To achieve this, the position of each ingredient in the pattern is remembered before calculation and then pattern is reconstructed with the items that are extracted from the internal storage when the crafting operation is performed. 

I was unable to make the pattern always exactly match the input pattern (non-exact patterns have very weird shapes) and as such the pattern is always reconstructed into a 3x3 grid with empty stacks where no items belong. As such I had to remove the size checks out of `CraftingPattern#getOutput` and `CraftingPattern#getByproducts`.

With that in place it was very easy to implement the same for processing pattern. Implementing #1937.

This does break existing crafting tasks. But I've put in some compatibility code that makes any old crafting task not run. They can simply be canceled in the crafting monitor. 

I've also removed some remnants of v5 that were not used anymore.


raoulvdberge:
I don't understand the problem / bug that you are solving.

raoulvdberge:
Can you refer to me in the original code where the problem is? For me, at the moment, this is just a large blob of code where I don't understand your reasoning and changes.

Darkere:
Gonna redo this to work with the refactored crafting... In the next week maybe.

PR #2625, opened 2020-07-17
Autocrafting refactor

raoulvdberge:


src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/calculator/CraftingCalculator.java
```diff
@@ -0,0 +1,362 @@
...
+                                    ICraftingPattern pattern,
+                                    CraftingPatternInputs inputs,
+                                    IStackList<FluidStack> fluidsToExtract,
+                                    ProcessingNode node) throws CraftingCalculatorException {
+        int ingredientNumber = -1;
+
+        for (CraftingPatternInputs.Ingredient<FluidStack> ingredient : inputs.getFluidIngredients()) {
+            ingredientNumber++;
+
+            PossibleFluidInputs possibleInputs = new PossibleFluidInputs(ingredient.getInputs());
```
Darkere, line 255: PossibleInputs and PossibleFluidInputs can be merged.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/PossibleInputs.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public void sort(IStackList<ItemStack> mutatedStorage, IStackList<ItemStack> results) {
+        possibilities.sort((a, b) -> {
+            ItemStack ar = mutatedStorage.get(a);
+            ItemStack br = mutatedStorage.get(b);
+
+            return (br == null ? 0 : br.getCount()) - (ar == null ? 0 : ar.getCount());
+        });
+
+        possibilities.sort((a, b) -> {
```
Darkere, line 42: I never quite understood why this calls sort with 2 lists instead of calling sort twice.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/calculator/CraftingCalculator.java
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+    }
+
+    private void calculateForFluids(int qty,
+                                    IStackList<ItemStack> storageSource,
+                                    IStackList<FluidStack> fluidStorageSource,
+                                    IStackList<ItemStack> results,
+                                    IStackList<FluidStack> fluidResults,
+                                    ICraftingPattern pattern,
```
Darkere, line 246: pattern is not used

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public final class IoUtil {
+    private static final Logger LOGGER = LogManager.getLogger(IoUtil.class);
+
+    private static final int DEFAULT_EXTRACT_FLAGS = IComparer.COMPARE_NBT;
+
+    public static IStackList<ItemStack> extractFromInternalItemStorage(IStackList<ItemStack> list, IStorageDisk<ItemStack> storage, Action action) {
```
Darkere, line 27: 1st parameter is a Collection

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
```diff
@@ -0,0 +1,187 @@
...
+                return null;
+            }
+
+            extracted.add(result);
+        }
+
+        return extracted;
+    }
+
+    public static IStackList<FluidStack> extractFromInternalFluidStorage(IStackList<FluidStack> list, IStorageDisk<FluidStack> storage, Action action) {
```
Darkere, line 47: 1st Parameter is a Collection not a Stacklist

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/CraftingManager.java
```diff
@@ -94,32 +91,30 @@ public void cancel(@Nullable UUID id) {
     }
 
     @Override
-    @Nullable
-    public ICraftingTask create(ItemStack stack, int quantity) {
```
Darkere, line 98: create methods can be easily combined

Darkere (commented):
About the fluid calculation. There was actually a bug in my side in there that probably appeared in one of the merges I did. 

I intentionally wanted to not allow fluid calculation to combine 2 different fluids for the same input. You could easily end up in a situation where the system will try to insert 100ml of one fluid and 900ml of a different fluid into a machine with a single tank.

The same issue is possible with items (although in that case I would argue it is much more rare). I've not found a good solution for this.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
raoulvdberge: I changed it so it takes a stacklist, since it was using `Collection<StackListEntry<?>>`, so if it enforces a stack list entry might as well just use the stacklist. Makes the signature easier to read.

Darkere: ah. Then you need to remove getStacks() from where this gets called though.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/CraftingPatternInputs.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    @Nullable
+    private Ingredient<ItemStack> findMatchingItemIngredient(NonNullList<ItemStack> inputsForSlot) {
+        for (Ingredient<ItemStack> existingIngredient : itemIngredients) {
+            if (existingIngredient.getInputs().size() == inputsForSlot.size()) {
+                for (int i = 0; i < inputsForSlot.size(); i++) {
+                    if (!API.instance().getComparer().isEqualNoQuantity(existingIngredient.getInputs().get(i), inputsForSlot.get(i))) {
+                        break;
```
Darkere, line 73: You are breaking out of the wrong loop here.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/calculator/CraftingCalculator.java
raoulvdberge: Not true, pattern is used

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
raoulvdberge: That is already done?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/CraftingManager.java
raoulvdberge: I don't see how. One is for ItemStack, one is for FluidStack? Also the getPattern calls are specific to the stack type.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/PossibleInputs.java
raoulvdberge: We want to first sort on the network storage, then sort on the items we have in the crafting calculation internal storage. But the sort calls could be combined, yeah.

raoulvdberge:
> I intentionally wanted to not allow fluid calculation to combine 2 different fluids for the same input. You could easily end up in a situation where the system will try to insert 100ml of one fluid and 900ml of a different fluid into a machine with a single tank.

But, wouldn't you change the crafting pattern for this? This can only happen if the user explictly marks 2 fluids to be equivalent to each other in the allowed tag list, right?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/calculator/CraftingCalculator.java
Darkere: where? I'm not finding it (and intellij isn't either).

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
Darkere: Then you did not commit those changes yet.

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/CraftingManager.java
Darkere: Oh. Nvm. You can combine most of the code into an extra method if you want to. 

Fully combining it requires making `CraftingManager#getPattern` and `CraftingRequestInfo` generic + a generic subclass.

raoulvdberge: I'll leave it as-is

Darkere:
> This can only happen if the user explictly marks 2 fluids to be equivalent to each other in the allowed tag list, right?

Correct. Imagine 2 mods adding an equivalent fluid under a different name. Pack/Mod authors will add compatibility by either adding a recipe for both or by changing the recipe to accept all fluid with that tag. 

The machine will accept either fluid. But it will most likely not convert one to the other, so that they could fit into the same tank.

raoulvdberge:
Yes, but that wouldn't happen as long as the user doesn't modify the allowed tag list, correct?

Darkere:
Indeed. Allowing that is however the point of that function right?

We are also not communicating to the player that this is the reason for it not working. He'll just see that the system can't insert the original fluid.

raoulvdberge:
As discussed on Discord, the current fluid processing impl is broken anyway. So let's merge this.

PR #2633, opened 2020-07-25
fix more ConstructorNetworkNode duping

Darkere:
Was SUCCESS is now CONSUME. (That means we are duping items again :D)

By checking both we save headaches with other mods not noticing the change (its also what vanilla uses). 

There are more places where it might be sensible to change from SUCCESS to CONSUME but as far as I can tell none are critical.


raoulvdberge:
Thx

PR #2673, opened 2020-09-11
Allow extending for all the blocks

Darkere:
I only "need" like half of them, but I figure making them all work like that is a bit cleaner. 

Currently, just copy-pasting almost the whole block class for these. It works. But it's quite ugly. 

I'm slowly getting close to release:
image

Also changed the NetworkCardItem to recognize anything extending NetworkReceiverBlock instead of just RS's NetworkReceiver.


Darkere:
Alright lets get this going. More PRs (that can stay open forever) \o/

PR #2703, opened 2020-10-10
[MC1.12] [Client] Performance improvements for crafting grid

ScoreUnder:
I was playing Direwolf20, and with a very diverse inventory in my RS system (many damaged rods and enchanted books that I hadn't yet voided, from an automatic fishing setup) plus tens of thousands of cobblestone, I decided to save space by compacting the cobblestone (3x3 cobblestone crafts into 1 compressed cobblestone). I noticed that every time I shift-clicked a stack of compressed cobblestone out of the crafting grid, my client froze for about half a second, which made bulk crafting a little difficult -- I could spam click, but that froze the client for even longer and could lead to a disconnection.

I looked into the issue with visualvm and the vast majority of the time was spent on sorting the grid view. These commits represent my adventure into that performance issue.

I have gone and tidied up a couple of things in the process, so we'll see how controversial that is.

Bugs I attempted to address:

- Crafter manager ORs terms rather than ANDing them
- No inventory is shown when filter term ends in `|` (e.g. `@ref|`, when about to type another term)

Performance improvements:

- Created one big chained comparator for the grid sort, rather than re-sorting multiple times
- Avoided re-sorting grid on delta update, instead opting for in-order reinsertion of affected stacks (quite easy to do with the single comparator)

Misc:

- Avoided lists of grid stack filters where possible, and instead joining them with a single filter that ANDs them together. This ties into the crafter manager bug; I noticed it during this refactor and decided it must have been a mistake.
- Merged `GridViewBase` subclasses into one (`GridViewImpl`). With a new `setQuantity` method on the grid stacks, they no longer need to be separate things.

I am considering also adapting this commit for 1.16, but I'd like to get this in for 1.12 just so that anyone else with this issue on direwolf20 specifically can update this mod to fix it.

Anyway, thanks for the great mod! It's easily one of my favourites in the pack.


ScoreUnder:
Do not merge yet

raoulvdberge:
I don't pull changes for 1.12. please only target 1.16.

raoulvdberge:
Shift clicking issues have been solved in 1.16 as well.

ScoreUnder:
I see. It feels like a shame to keep the direwolf20 drop-in replacement version to myself.

I'll see about porting this up to 1.16 later today (need to get groceries before everything shuts). The single comparator rather than multiple-sort approach seems like a performance improvement that is still applicable to 1.16, so would you like me to try porting the whole thing over? Or are there some parts which are less desirable?

Darkere:
Shift clicking Was still slower than I wanted it to be. I adressed extraction which only really was a problem with loads of lookups in external storage. 

If its the sorting that is the problem then simpler solution could be to prevent sorting ( I explicitly filtered shift clicking on the crafting output.

raoulvdberge:
Please port over the entire thing, then I can see what we keep and what we don't :)

ScoreUnder:
Roger 🙂

2020年10月10日(土) 15:01 Raoul <notifications@github.com>:

tth05:
@ScoreUnder I'm maintaining a fork for 1.12. You can PR these changes there if you want to you! Otherwise, would it be ok for you if I port these over?

ScoreUnder:
Please do!

I'm currently working on redoing this for 1.16 -- or should I say, I'm currently working on trying to find creative ways to fill a RS system up with junk in the absence of all my favourite junk-item-generating mods. 😅

If you need me to make a PR to your repo I can, otherwise feel free to just fetch and merge (assuming it doesn't throw conflicts out everywhere)

tth05:
Alright thanks!

P.S. I recommend just generating the items from code with random NBT data. That's an easy way

PR #2705, opened 2020-10-10
[MC1.16] [Client] Performance improvements for grid view

ScoreUnder:
See #2703. The changes are roughly the same as described there.

Testing this (and specifically, trying to reproduce the performance problems) was probably the hardest part. I stuck a bunch of importers around a chest and used a command block to fill it with junk:

    loot insert 150 63 48 fish minecraft:gameplay/fishing/treasure 150 62 48

This got me a lot of non-stacking items, and it was just a matter of leaving this running for a while.

To elaborate on the bugs I mentioned in the previous PR:

----

Crafter manager search bug:

1. Add some patterns to your RS system: grid, pattern grid, pattern, iron block
2. Open your crafter manager
3. Search for "pat gr"

Expected result: pattern grid shows up only

Actual result: pattern, grid, and pattern grid all show up as if you had typed `pat|gr` (which also does the same)

----

`|` at end of search bug:

1. Have many items in RS
2. Type `@ref|`

Expected result: Since the user has (probably) not finished typing, the `@ref` search should still be valid. Either that, or it should be equivalent to "`@ref` or `<blank>`", i.e. all items.

Actual result: Grid is blank

----

For completeness, screenshots (see perf graph on bottom left):

1. [No RS grid open](https://i.imgur.com/qZzoX8S.png)
2. [RS grid open without sorting changes](https://i.imgur.com/v1QwePO.png)
3. [RS grid open with sorting changes](https://i.imgur.com/44Mz1AV.png)

The regular lag spikes in 2 are from the importers constantly pulling in junk from the chest, and the grid view having to keep up with that. In 3 it looks like they're gone altogether (I did visually confirm that items were still jumping around in the grid). Note that frame times on the right-hand half of the graphs might be a little inflated because my screenshot utility is drawing above minecraft.

With this change, `preventSortingWhileShiftIsDown` can be safely disabled with negligible performance impact. Now it's just an option affecting gameplay, instead of performance overall, and can be enabled by anyone who doesn't like items jumping around under their cursor when they shift+click them.


Darkere:
Ah you misunderstood what I meant :D . 
The grid screen has a specific function for NOT preventing sorting when shift crafting. Removing that would have prevented sorting when shift crafting and potentially improved performance. 

Your solution is much better ofc :D

ScoreUnder:
Just noticed while playing around that this fixes another problem... The search `#:` is pretty slow (for obvious reasons I guess), so before this change if anything is actively importing/exporting in a large inventory, depending on how large the inventory is and how fast the imports/exports are happening, it has the potential to completely freeze the client as it can't keep up with changes from the server. With this change, after the initial lag spike, that search is mostly fine.

raoulvdberge:
This looks cool! I will try to review this PR this weekend.

ScoreUnder:
Rebased to fix merge conflicts

ScoreUnder:
Rebased again, same reason

raoulvdberge:
Thanks!

PR #2706, opened 2020-10-10
[Server] Speed up chunk loading when many similar items in RS

ScoreUnder:
> When there are similar items (same type, different enchantments and
damage) in an RS system, it can sometimes take a very long time to load.
> 
> This should take a stab at fixing that, with the caveat that it
potentially reduces the flexibility of `IStackList#get` and
`IStackList#getEntry` (now forces at least COMPARE_NBT, though as far as
I can tell there was never a time when this wasn't the case already).

When creating a system saturated with junk for #2705, I ran into the problem that my singleplayer world was taking a while to load (about 1m 17s by the end of it). Here's something that takes that same world down to 12s from first click to landing in the world.

The junk tested with was fishing "treasure", so a lot of bows, fishing rods, and enchanted books. The fishing rods and bows all have varying damage levels, and everything has varying enchantments.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/api/util/IStackList.java
+++ b/src/main/java/com/refinedmods/refinedstorage/api/util/IStackList.java
@@ -76,15 +76,24 @@ public interface IStackList<T> {
     /**
      * Returns a stack.
      *
+     * {@code flags} is assumed to contain at least {@code COMPARE_NBT}, because
+     * otherwise there may be multiple candidates to return.
+     *
      * @param stack the stack to search for
      * @param flags the flags to compare on, see {@link IComparer}
      * @return the stack, or null if no stack was found
      */
     @Nullable
-    T get(@Nonnull T stack, int flags);
+    default T get(@Nonnull T stack, int flags) {
+        StackListEntry<T> entry = getEntry(stack, flags);
+        return entry == null ? null : entry.getStack();
+    }
 
     /**
      * Returns a stack entry.
+     *
+     * {@code flags} is assumed to contain at least {@code COMPARE_NBT}, because
+     * otherwise there may be multiple candidates to return.
      *
      * @param stack the stack to search for
      * @param flags the flags to compare on, see {@link IComparer}
--- a/src/main/java/com/refinedmods/refinedstorage/api/util/IStackList.java
+++ b/src/main/java/com/refinedmods/refinedstorage/api/util/IStackList.java
@@ -76,9 +76,6 @@ public interface IStackList<T> {
     /**
      * Returns a stack.
      *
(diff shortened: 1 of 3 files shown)


raoulvdberge:
How saturated was your system? Was it reastical? I don't think we need to overoptimize if it doesn't really happen in the real world.

Darkere:
People storing all their junk from mob farms is unfortunately not uncommon :/ Same with running fishing setups for forever. 

The mega disks that are so commonly used also don't discourage that kind of behavior (although they do help with performance).

raoulvdberge:
I believe this is a good solution, but I don't like the fact that we're accepting flags and then silently ignoring the nbt comparison.

Can we remove the `flags` parameter and force strict NBT checks at all times?

tth05:
Wouldn't this break for example exporters with compare NBT off. Exporting for example all enchanted books wouldn't work anymore

raoulvdberge:
This only affects the stack list. Nbt insensitive extractions are done through the disk iirc.

tth05:
Oh yeah that's true. A better solution might be to write a HashSetMultiMap and add equals and hashcode to StackListEntries. Then the second access would also be O(1).
This would speed up everything and still allow for non-exact extraction and it could be used in the StorageCache as well.

ScoreUnder:
> Then the second access would also be O(1).

Can you expand a little? This one went over my head

raoulvdberge:
Which second access?

tth05:
I mean the second access on the collection after accessing the MultiMap using the Item instance. The first one returns the ArrayList from the MultiMap and then an O(n) search is performed which is the thing that slows everything down if you have lots of ItemStacks of the same Item. 

What I mean is basically implementing this PR, but instead of on the MultiMap itself apply it to the Collections stored within the MultiMap. Then you'd access the MultiMap first by Item and then the result by ItemStack, but it leaves the ability to access without checking for NBT data because all ItemStacks are still indexed using their Item.

ScoreUnder:
That makes sense!

I think you're right about ItemStackList eventually being used by exporters/importers, so I'll take your suggestion to keep the multimap.

ScoreUnder:
It came out a little differently than described while I was forming the idea in code but I'm pretty sure it still makes sense this way.

I've tested with importing/exporting enchanted books with and without "exact mode", and it behaves as expected.

raoulvdberge:
I'm going to start writing some basic unit tests for the stack list, based on how the public API works now.

Then, I'm going to merge this PR and then we'll be sure everything works as intended.

ScoreUnder:
Rebased to fix merge conflicts. Seems to pass the tests that have been added so far

PR #2707, opened 2020-10-11
2-way JEI sync on search boxes

ScoreUnder:
Lets you type in either search box (when JEI sync is on) and lets you use JEI's searchbox as a "memory" between closing and reopening the GUI.

If I had to guess, the feedback on this is probably going to be "needs toggleable option" and I'd be happy to add that if that's what you're aiming for.


raoulvdberge:
Cool beans! Can you add two new Grid searchbox options?
- JEI synchronized (two-way)
- JEI synchronized (two-way autoselected)

ScoreUnder:
Looks like it will need new translation keys

ScoreUnder:
Rebased to resolve conflicts

raoulvdberge:
Very nice work, clean code 👍

PR #2708, opened 2020-10-15
Fix crash on shift scroll up on servers

Darkere:
You probably got the crash report from FC

getSlotFor is clientside only -.-


raoulvdberge:
Fixes #2709

raoulvdberge:
Changelog entry please!

PR #2710, opened 2020-10-15
Fix things working without power

Darkere:
Wow that was a lot more than I thought D:


src/main/java/com/refinedmods/refinedstorage/apiimpl/network/node/GridNetworkNode.java
```diff
@@ -439,6 +439,16 @@ public void onClear(PlayerEntity player) {
                     network.getItemStorageTracker().changed(player, slot.copy());
                 }
             }
+        } else if (network == null || !network.canRun()) {
```
raoulvdberge, line 442: I think this will dupe items if you have a pattern grid? Can you improve this whole if so it's readable?

```java
if (type == GridType.CRAFTING) {
  if (network != null && network.canRun() && network.getSecurityManager().hasPermission(Permission.INSERT, player)) {
    // stuff
   } else {
    // other stuff
   }
} else if (type == GridType.PATTERN) {
  clearMatrix();
}
```

PR #2729, opened 2020-10-30
Add the ability to calculate in the CraftingSettingsScreen.java

danieldietzler:
Add the ability to enter mathematical expressions in the amountField.


raoulvdberge:
It doesn't really feel right to accept arbitrary Js expressions. Besides, the Js engine is removed in later java versions.

Daniel-Dietzler:
So would you say there's a point on building a complete parser with computing laws etc? Or is it just to minor/irrelevant to put more effort into it?

Daniel-Dietzler:
Apparently there is an open source project called "EvalEx" on GitHub which could be used.

raoulvdberge:
I made a parser recently. We could include that.

Daniel-Dietzler:
I'd appreciate it because often when you're building big multiblocks you need to have like 15x15x17 blocks or something. In those cases just typing 15x15x17 would be a minor but really neat feature.

raoulvdberge:
I'll close this for now.

PR #2739, opened 2020-11-13
Update ja_jp.json

alyxandraio:
Just some general translation fixes. Makes the style more informal as opposed to formal. Changes terms and nouns to those more consistent with Minecraft's official Japanese translation to minimize confusion.


raoulvdberge:
Thank you

PR #2786, opened 2020-12-28
cache filled disks fixes #2771

### LINKED ISSUE #2771: Lag from inserting into many full disk/full system
#### Issue description:
When importing into a full system. TPS dies. 

#### Steps to reproduce:
1. System with 20+ full disks
2. insert into the system quickly

- Minecraft:1.16.4
- Forge: 35.1.0
- Refined Storage:1.9.9

image
https://spark.lucko.me/#Bp0xk1aXZ5

Not sure if we can maybe cache disks that are completely filled and not evaluate them. Otherwise just caching getStored() should already yield significant improvements.

Darkere:
~100 filled disks and went from additional +140ms down to +2ms. 

Still not ideal but at least it's not that bad anymore.

Could still throttle the disk manipulator but that wouldn't affect normal insertion.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/ItemStorageDisk.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/disk/ItemStorageDisk.java
@@ -35,6 +35,7 @@ public class ItemStorageDisk implements IStorageDisk<ItemStack> {
     private final int capacity;
     private final Multimap<Item, ItemStack> stacks = ArrayListMultimap.create();
     private final UUID owner;
+    private boolean isFull;
 
     @Nullable
     private IStorageDiskListener listener;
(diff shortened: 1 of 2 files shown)


raoulvdberge:
Maybe it's a better idea to cache/only update the result of `getStored()` when needed... And then just compare `stored == capacity` to check whether we're full capacity. Then the stored count is cached as well.

Darkere:
Yep that's a good idea. `getStored()` does get called a bunch.

raoulvdberge:
Can't we keep an integer and just increment/decrement that instead of recalculating everytime?

tth05:
That's what I did in fork as well. And `getStored()` uses streams which in itself makes it A LOT slower already.

Darkere:
Sure can do. Just need to actually look at the code now 😄

raoulvdberge:
Cool, can we do this for fluids as well? For consistency.

Issue #2631, opened 2020-07-25
Can't give or take items to RS. When Immersive Engineering Fluid Tank is connected as External Storage


archer9234:
#### Issue description:
Unable to remove or put items into RS.

#### What happens:
When attaching external fluid storage to Immersive Engineering's Tank. Attempting to take items in/out of the grid. Either from other external sources or RS storage, fails. 

#### What you expected to happen:
Fluid Grid connected to Immersive Engineering doesn't block interaction with RS.

#### Steps to reproduce:
1. Make IE Fluid Tank. Not sure if it's all fluid's. But use Creo oil
2. Connect External Storage to tank.
3. Attempt to add or take items out of any connected storage to RS, will not work. Grid items will also remain visible when fully disconnected from the entire system.
...

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.16.1
- Forge: 32.0.69
- Refined Storage: 1.9.2

Does this issue occur on a server? [yes/no] Yes

#### If a (crash)log is relevant for this issue, link it here:

[pastebin/gist/etc link here]

Edivad99:
I think it is the same problem reported here #2568

Darkere:
Duplicate of #2568

amohacker:
I also have this bug in 1.16.5 specifically when i have an immersive engineering fluid barrel attached to the system i become unable to take items out of the system after i put an item into the system.
The other fluid containers seem to work fine it's just the barrel as far as i can tell

PR #2789, opened 2020-12-28
fix blockupdates always invalidating storage fixes #2568 #2631

### LINKED ISSUE #2568: [1.15.2] Can't remove items from RS grid when connected to DimStorage chest
#### Issue description:
When I connect a dimstorage chest to my RS system using the external storage I can't take any items from my RS grid. Not just the items from the dimstorage chest. I already reported it on the dimstorage github but they told me I should contact you: https://github.com/Edivad99/DimStorage/issues/27

#### Steps to reproduce:
1. Connect a dimstorage chest to an RS system using external storage. 
2. Try to remove items from the RS system (not only the items from the chest but also from drives). If you're fast enough you might be able to take something once but not after that.

#### Versions:

- Minecraft: 1.15.2
- Forge: 31.2.9
- Refined Storage: 1.8.4
- DimStorage: 3.1.5

Yes

Darkere:
This should fix issues with being unable to extract items from the grid. 
Hopeful this doesn't have any side effects but not can't really guarantee that. 

Can't really think of a case where this fails though.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/node/ExternalStorageNetworkNode.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/node/ExternalStorageNetworkNode.java
@@ -22,6 +22,7 @@ import com.refinedmods.refinedstorage.tile.ExternalStorageTile;
 import com.refinedmods.refinedstorage.tile.config.*;
 import com.refinedmods.refinedstorage.util.AccessTypeUtils;
 import com.refinedmods.refinedstorage.util.StackUtils;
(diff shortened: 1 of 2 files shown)


raoulvdberge:
What if you just open a world with the external storage and connected block present? Then no block update is fired, which means it would still break on the first update.

raoulvdberge:
What I mean is that `connectedBlockState ` is only setup right after the update request. But it isn't set when the world loads, causing a invalidation anyway for the first time.

Darkere:
Now I understood :)

raoulvdberge:

PR #2809, opened 2021-01-04
Pass "compare" flags into extractItem/extractFluid in more places

ScoreUnder:
The "exact" button wasn't working for me (behaved as if always on) when dropping items from a constructor. This should fix that, and hopefully cover the remaining cases where the comparison type isn't properly propagated down to later functions.


raoulvdberge:
Can  you add a changelog entry

Issue #2811, opened 2021-01-05
crash


AdnanuReeves:
---- Minecraft Crash Report ----
// Why did you do that?

Time: 1/5/21 5:09 PM
Description: Ticking screen

java.lang.UnsupportedOperationException: null
	at java.util.AbstractList.add(Unknown Source) ~[?:1.8.0_271] {}
	at com.refinedmods.refinedstorage.screen.grid.view.GridViewImpl.addStack(GridViewImpl.java:181) ~[refinedstorage:1.9.11] {re:classloading}
	at com.refinedmods.refinedstorage.screen.grid.view.GridViewImpl.postChange(GridViewImpl.java:170) ~[refinedstorage:1.9.11] {re:classloading}
	at com.refinedmods.refinedstorage.network.grid.GridItemDeltaMessage.lambda$null$0(GridItemDeltaMessage.java:65) ~[refinedstorage:1.9.11] {re:classloading}
	at java.lang.Iterable.forEach(Unknown Source) ~[?:1.8.0_271] {}
	at com.refinedmods.refinedstorage.network.grid.GridItemDeltaMessage.lambda$handle$1(GridItemDeltaMessage.java:65) ~[refinedstorage:1.9.11] {re:classloading}
	at com.refinedmods.refinedstorage.screen.BaseScreen.runActions(BaseScreen.java:78) ~[refinedstorage:1.9.11] {re:classloading}
	at com.refinedmods.refinedstorage.screen.BaseScreen.runActions(BaseScreen.java:68) ~[refinedstorage:1.9.11] {re:classloading}
	at com.refinedmods.refinedstorage.screen.BaseScreen.func_231023_e_(BaseScreen.java:118) ~[refinedstorage:1.9.11] {re:classloading}
[...]

PR #2812, opened 2021-01-05
Crash #2811 fix

ScoreUnder:
Only the first commit is necessary to avoid the crash. The second is to avoid updating the inventory when it shouldn't, and the third is to safeguard against technically-non-breaking Java implementation changes.

→#2811


raoulvdberge:
Can you add a changelog entry?

PR #2817, opened 2021-01-08
Add all blocks to loot generator

MelanX:
This PR adds the remaining handwritten blocks to loot table generator. 
It makes the BLOCKS field in RSBlocks private -> public, I hope it's not a big problem for you.


raoulvdberge:
Why?

MelanX:
If you add anything else, you don't need to write a new block loot table. You can simply use runData. And if anything changes in future, you can re-run it and the new generated loot tables are created.

raoulvdberge:
I appreciate the effort, but I don't think this is needed.

PR #2825, opened 2021-01-12
Update Disk Drives UUID when in the Grid Network.

KewaiiGamer:
When auto crafting disk drives they do not get their UUID added until they go to a player inventory. Without a UUID it is impossible to add the Storate Drive to a Disk Drive with a hopper or a Refined Pipe since they never go to a player inventory

[https://cdn.discordapp.com/attachments/788122134136094722/798538134383493150/QWi6rmzoqI.mp4](Video of Storage Drives not being inserte into Disk Drives without having an UUID)

This is very necessary since this issue prevents fully automating expanding storage


raoulvdberge:
Updating it in the network node seems like the wrong place. As well as using the storage cache seems wrong.

Darkere:
There should be a way to override ItemStack creation and add the tag there.

This would have the added benefit of being able to copy disks in creative and avoids "duped" disks from linking to the same inventory. 

PS: I remember that disk drives cannot be inserted into through automation... Although I could be wrong about that.

PR #2858, opened 2021-01-30
fix fluid conversion in in IOUtil fixes #2772

### LINKED ISSUE #2772: 1.16.4 fluid autocrafting duplication bug
#### Issue description:
fluid autocrafting 'duplicates' fluids stored both internally and externally when autocrafting by raising the 'stored' value of the fluid till autocrafting completes.

#### What happens:
when autocrafting with fluids, as soon as autocrafting begins, the 'stored' value of the fluid gradually goes up till it reaches a maximum ( which seems to be related to the amount of fluid storage present in the system ) or the autocrafting is complete. When autocrafting is done, the 'stored' value of the fluid remains, the 'crafting' won't complete until canceled, and all fluid containers, both fluid storage drives and external tanks, are filled with the 'stored' fluid, thusly duplicating it. 

When autocrafting with fluid, X number of mB are either stored or processed, used up, and autocrafting completes when processing does.

#### Steps to reproduce:
1. create recipe that used fluid autocrafting (ie 1 empty bucket, 1 B lava, in a tank, to produce 1 lava bucket)
2. craft with said recipe, from one bucket to many
3. receive infinite fluid
...
(Note this seems to happen only on bigger systems, I tried it in a fresh world with a minimal RS system, and I had no issues. This has happened to me on RS 1.9.8 on MC 1.16.3 as well, both times it only occurred when my RS system got larger.)

- Minecraft: 1.16.4
- Forge: 35.1.13
- Refined Storage: 1.9.9

yes ( tested on lan )

no logs nor crashes are reported during the issue, however I can produce a  video if need be.

Screenshots:
Autocrafting Setup:
autocrafting_setup

Autocraft Recipe:
autocraft_recipe

Tank Initial Amount:
tank_initial_amount

Autocraft Mid-craft:
autocraft_mid_craft

autocraft finished:
autocraft_finished

tank after autocrafting and autocraft is cancelled:
tank_autocraft_finished

Darkere:
fix fluids being transformed if another mod gives back the wrong fluid

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/IoUtil.java
@@ -172,7 +172,7 @@ public final class IoUtil {
             FluidStack result = network.extractFluid(toExtract.getStack(), toExtract.getStack().getAmount(), Action.PERFORM);
 
             if (!result.isEmpty()) {
-                internalStorage.insert(toExtract.getStack(), result.getAmount(), Action.PERFORM);
+                internalStorage.insert(result, result.getAmount(), Action.PERFORM);
 
                 toRemove.add(result);
             }


raoulvdberge:
Can you add a changelog entry

PR #2873, opened 2021-02-07
fix interface duping issue

### LINKED ISSUE #2849: Interface duping/flooding network under specific conditions
#### Issue description:

When an interface (without a stack upgrade) inserts items to a network with at least one full storage disk, the interface starts duping items in the next disk.

#### What happens:

Specifically, the interface only loses one item per operation as normal, but the network gains the entire contents of the first stack in the interface every operation.

Network receiving only one item per operation, i.e., the items that disappear from the interface.

#### Steps to reproduce:
1. Place a creative controller, disk drive, grid, and interface down as part of the same network.
2. Put two 1k storage disks in the drive and manually fill the first disk to (for example) 980/1000.
3. Put several stacks of items in the interface.
4. Watch the disk drive usage. It goes up to 1k normally (in increments of 1) then starts filling up much more quickly.

- Minecraft: 1.16.4
- Forge: 35.1.37
- Refined Storage: 1.9.11

Does this issue occur on a server? [Tested only in SSP, don't have a server environment to test right now.]

Darkere:
ofc it was entirely my fault :/ Did not realize the stack HAD to be copied.

Still don't see the point in getCacheDelta but at least it is not broken :)

fixes #2849


raoulvdberge:
Ah, no worries

PR #2881, opened 2021-02-17
Ported covers from 1.12

Buuz135:
Covers need forge's experimental lighting to be enabled to have proper lighting

Currently missing features/Issues:

- [x] Cover collision and bounding boxes
- [x] Implementation for other things outside cables
- [x] Config option to hide covers from JEI
- [x] Cover recipes
- [x] When placing the cover in a north face it doesn't get placed but placing one on the bottom side it will place both
- [x] Improve cover syncing code if possible
- [x] Cover/Item connectivity interaction
- [x] Better block checking for possibe covers (As most methods now require a world)
- [x] Cover names
- [x] Better texture based on cover sideness

More issues to come (Probably)


Darkere (changes requested):
Few things to do :)
honestly a bit concerned the recipe will come back to bite us... but oh well I guess that is a problem for pack devs

PR #2895, opened 2021-03-01
Revert ja_jp.json

314systems:
I revert ja_jp.json because it has a lot of unnatural or incorrect translations.


raoulvdberge:
There is a lot of discussion around the Japanese translation. I can keep merging PRs like these, but this translation keeps shuffling around.

Issue #2768, opened 2020-12-11
Z fighting on stock count at the Storage Monitor


NeverSnows:
#### Steps to reproduce:
1. Craft Storage Monitor;
2. Display an item.

#### Versions:

- Minecraft: 1.16.3
- Forge: 1.16.3 - forge - 34.1.24
- Refined Storage: 1.9.8

This issue occurs on Servers and single player mode.

PR #2933, opened 2021-04-02
Fix flickering issue with storage monitors.

tivec:
This PR will add a *tiny* little offset to the position of the text, getting rid of the z-fighting. This fixes issue #2768.

Note: The text will be offset both in X and Z direction and does not care about the actual orientation of the block. I'm unfortunately not well versed enough with forge to figure out which side is the front. However, the difference is so minuscule that it shouldn't be noticeable.


raoulvdberge:
Thank you! I incorporated your fix in 884610100eb4c949a34c291ce747a21440cbd8ae

PR #2974, opened 2021-05-12
fix detector crashing when dyed and deactivating powered state

### LINKED ISSUE #2899: [1.16.5] Crash when Detector dyed in-world
#### Issue description: 
Right-clicking on a detector with a dye will crash the game, whether or not it is currently outputting a signal and lit up. Dying will still work properly when crafting the dye with the detector.

#### What happens: 
The game crashes when you right-click on a detector in-world with a dye.

The detector to be dyed to the relevant colour.

#### Steps to reproduce:
1. Place Detector of default colour
2. Right-click with dye (my tests have been specifically Magenta Dye)
3. Game will crash

- Minecraft: 1.16.5
- Forge: 36.0.14
- Refined Storage: 1.9.12

Does this issue occur on a server? [yes/no] Unknown

https://pastebin.com/H1Xqw7gE

Darkere:
fixes #2899


raoulvdberge:
Why are we no longer using `changeBlockColor`?

raoulvdberge:
I'd much rather have that we change `changeBlockColor`  and not add a weird exception here.

Darkere:
I realize midway through that the detector has its own "powered" state, and no connected state. That's why those are not in the Colormap class. Same as the controller.

Issue #2921, opened 2021-03-25
Suggestion: try use items you have the most of when adding from JEI


Rodg88:
#### Issue description:
When you insert a recipe from JEI, the items that get moved to the grid are sometimes not what you have the most of (eg defaulting to crying obsidian over plain obsidian, or if you're in a mod pack with custom tags, it might put items from a tag that don't make sense to the user. (example in Enigmatica 6, it would always try use treated sticks before sticks). Can be more annoying when making autocrafting patterns especially.

I think the ideal solution would be to try use the item you have the most of first. But vanilla items prioritization could also work, except for crying obsidian and maybe wood types etc.

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.16.5
- Forge: 36.1.0
- Refined Storage: 1.9.12

PR #2975, opened 2021-05-13
JEI transfer picks the most common ingredient

Darkere:
When transfering items via JEI it will look up the most common item in the system and use that one instead of randomly picking. 
Implements #2921

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/CraftingGridBehavior.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/CraftingGridBehavior.java
@@ -1,6 +1,7 @@
 package com.refinedmods.refinedstorage.apiimpl.network.grid;
 
 import com.refinedmods.refinedstorage.api.autocrafting.ICraftingPattern;
+import com.refinedmods.refinedstorage.api.autocrafting.ICraftingPatternProvider;
 import com.refinedmods.refinedstorage.api.network.INetwork;
 import com.refinedmods.refinedstorage.api.network.grid.GridType;
 import com.refinedmods.refinedstorage.api.network.grid.ICraftingGridBehavior;
@@ -286,7 +287,7 @@ public class CraftingGridBehavior implements ICraftingGridBehavior {
         for (int j = 0; j < player.inventory.getSizeInventory(); j++) {
             ItemStack inventoryStack = player.inventory.getStackInSlot(j);
 
-            if (inventoryStack.getItem() instanceof PatternItem) {
+            if (inventoryStack.getItem() instanceof ICraftingPatternProvider) {
                 ICraftingPattern pattern = PatternItem.fromCache(network.getWorld(), inventoryStack);
                 if (pattern.isValid()) {
                     for (ItemStack stack : pattern.getOutputs()) {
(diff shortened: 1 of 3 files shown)


DarkHelmet433:
This would be really, really helpful!

necauqua:
I played with this and my feedback is that I would also check for items that the system knows how to craft as a fallback - so if it has none of the variants it also looks up if it can craft one of them, this would help making patterns so much.

My existing scenario - making a pattern for a thing that needs blue dye, and I have a pattern to make vanilla dye from lapis, but of course I have none in the system.
And then there is a choice between vanilla blue dye that my system knows how to craft and some weird modded version that who knows how you need to even get that, and _of course_ the random selection picks up that weird modded version

But at least it stopped putting dark glass in all the recipes (if I have normal glass in the system), hurray

necauqua:
One more thing to make it just perfect - also check player inventory for patterns, so that when player creates a whole pattern chain with JEI (very, very common thing in modpacks) it pick up the right variants all the time even if the system does not know about such items yet

Darkere:
You know what... that is a great idea. Gonna blow peoples minds :D

necauqua:
I think it should prefer player patterns (like players rarely carry patterns, and when they do it's exactly because they are making chain recipes) over most items in system.

I also see that you included player items as well, this one (separately from player patterns) should not take preference over the system maybe

raoulvdberge:
> it picks up the right variants all the time even if the system does not know about such items yet

Doesn't this only work when you have none of all variants in the system?

If you have more of an item in your system than the pattern provides, the pattern would still be ignored.

necauqua:
Well that's exactly what I am talking about in my previous message

Darkere:
If I'm making patterns and there is an item that already exists in the system, I think it makes more sense to use that one rather than the item from a pattern that randomly can also be used in this recipe. 

More often than not I'll have plenty of these Items in the system and the recipe should use those rather than using the pattern for the special case I just made. 

For a concrete example, imagine you just made a pattern for Stone Bricks. I would still want any further patterns to still use the regular stone from the system. In fact it would be pretty weird if the machine casing recipe I make after this uses stone bricks rather than stone.

Darkere:
working nicely now :)

necauqua:
I got this btw:
<details>
<summary>exception</summary>

```
java.lang.UnsupportedOperationException: Crafting pattern is invalid
	at com.refinedmods.refinedstorage.apiimpl.autocrafting.InvalidCraftingPattern.getOutputs(InvalidCraftingPattern.java:58) ~[refinedstorage:1.9.13] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.IngredientTracker.addAvailableStack(IngredientTracker.java:39) ~[refinedstorage:1.9.13] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.createTracker(GridRecipeTransferHandler.java:112) ~[refinedstorage:1.9.13] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipeForCraftingGrid(GridRecipeTransferHandler.java:62) ~[refinedstorage:1.9.13] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipe(GridRecipeTransferHandler.java:53) ~[refinedstorage:1.9.13] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipe(GridRecipeTransferHandler.java:29) ~[refinedstorage:1.9.13] {re:classloading}
	at mezz.jei.transfer.RecipeTransferUtil.transferRecipe(RecipeTransferUtil.java:61) ~[jei:7.7.0.100]
```
</details>

Darkere:
oh yeah good catch. fixed.

NillerMedDild:
This feature would be greatly appreciated by many players, I hope you can find time at some point to review/merge this :x

raoulvdberge:
Hi, I'm currently on vacation. I'll take a look at this next week.

PR #2977, opened 2021-05-14
Add packet splitting for some packets

Darkere:
Wrote this for my own mod. It is not quite as nice as I would want it, but it works and is super trivial to adopt.


clienthax:
Is this going to get any love any time soon?, Been unable to play a pack for a while due to shoving too many nbt heavy items into my storage

KnownAsPower:
This should seriously be considered and added, multiple people are having issues with packets being too big and it seems like this would solve almost all if not actually all those issues. +1.

AlexHailstone:
Just adding a +1 from myself as well. I just ran into this issue shortly after a mob grinder and didn't know this was an existing issue. This would be super helpful especially knowing that there's the Extra Disks mod that allows up to 1billion items per disk and I've only gotten to around 3million items ish total so far. (obv this is due to the Armor from the mob grinder)

PR #3000, opened 2021-06-05
Improve performance for fast crafters

necauqua:
If a third-party crafter returns some stupid value from `getMaximumSuccessfulCraftingUpdates()`, then ~profiling~ sampling shows that `NodeRequirements.getSingleItemRequirementSet` is called a bunch of times and each time it creates ItemStacks which in turn gather capabilities-blah-blah, so it has a performance hit.

I observed that when using creative crafters while playing atm6 skyblock, they have a default value for getMaximumSuccessfulCraftingUpdates of 60k(!!!) and completely destroy server TPS when used for processing recipes.

This fix helps partially, it impoves performance for when the internal storage is small, but for deeply nested recipes the `IoUtil.extractFromInternalItemStorage` simulation is now hit, and I don't think I can fix that.

I decided to still make a PR for it so that just generally crafter performance is improved slightly.

P.S. 
Yeah for creative crafters obvious fix is to chage the config to something more sensible than 60k, but the big number works fine for crafting recipes, unless you request like millions of things, because the crafting node fulfills the request and then breaks out of the loop.
But unlike that, the processing node just keeps looping until the result is inserted into the network, this maybe can be looked at.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/NodeRequirements.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/NodeRequirements.java
@@ -12,6 +12,7 @@ import net.minecraft.nbt.ListNBT;
 import net.minecraftforge.common.util.Constants;
 import net.minecraftforge.fluids.FluidStack;
 
+import javax.annotation.Nullable;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Darkere:
60k??? lol some people...

This will help, although only very slightly.
It shouldn't matter what people have set for `maximumSuccessfulCraftingUpdates` as the loop gets cancelled the first time it is unable to insert its items. 

What I suspect you are looking at in that profile is some addon that is spawning an insane amount of Crafting Tasks?

necauqua:
If `hasAllRequirements` is false (because of them being crafted or idk yet the infrastructure of your code enough) it just goes through the 60k loop completely ruining the server, thats what happening for me

Darkere:
You are correct. There is a case where that can happen. Will fix.

Darkere:
would be amazing if you could test #3001

necauqua:
I can test in like an hour or two, irl happened

necauqua:
Yes, everything works fine with config back at 60k, sorry for the delay

raoulvdberge:
Thanks!

starforcraft:
> 60k??? lol some people...
> 
> This will help, although only very slightly.
> It shouldn't matter what people have set for `maximumSuccessfulCraftingUpdates` as the loop gets cancelled the first time it is unable to insert its items.
> 
> What I suspect you are looking at in that profile is some addon that is spawning an insane amount of Crafting Tasks?
Woops, I think I exaggerated a bit ;) lol

PR #3001, opened 2021-06-05
also break loop if we do not yet have enough ingredients

Darkere:
see #3000. We were running through all crafting updates in cases where the required items were not yet available.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
@@ -136,7 +136,7 @@ public class ProcessingNode extends Node {
                         break;
                     }
 
-                    if (hasAllRequirements && !canInsertFullAmount) {
+                    if (!canInsertFullAmount) {
                         if (allRejected) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
```diff
@@ -132,6 +132,8 @@ public void update(INetwork network, int ticks, NodeList nodes, IStorageDisk<Ite
                         if (canInsertFullAmount) {
                             canInsertFullAmount = IoUtil.insertIntoInventory(container.getConnectedFluidInventory(), extractedFluids.getStacks(), Action.SIMULATE);
                         }
+                    } else {
+                        break;
                     }
 
                     if (hasAllRequirements && !canInsertFullAmount) {
```
necauqua, line 139: Btw now `hasAllRequirements` is always true here (and can be removed) and the next if can be removed entirely (not removed but replaced with its content), since both `hasAllRequirements` and `canInsertFullAmount` are always true at that point)

necauqua:
Duplicating my comment from #3000 here so it's more into the context - this does indeed fix crafter performance, especially for creative crafters which have the default value of 60k crafting updates.

raoulvdberge:
Thanks!

PR #3002, opened 2021-06-06
Remove calculation threads causing concurrency issues

### LINKED ISSUE #2995: When auto-crafting, clicking on the Start button makes it become in a disabled state, but nothing happens after that.
#### Issue description: 
When auto-crafting, clicking on the Start button makes it become in a disabled state, but nothing happens after that. The GUI is just stuck there and nothing actually happens. I can click on the Cancel button to close it. When I check in the Crafting Monitor, nothing is being crafted nor is there the requested item in the Storage. Similar to a relatively old issue I found while trying to find a fix (#1908) however it's not exactly the same problem since I also checked for this same mod that caused the problem on that issue and it is not present in that modpack (the modpack is ATM6 - To The Sky)

#### What happens:
When initializing a crafting job the Start Button turns into disabled state and the crafting doesn't start.

image_2021-06-01_053156

The pre-crafting GUI opens and shows me what items / blocks are missing or the auto-crafting just starts.

#### Steps to reproduce:
1. Set up a RS system.
2. Try to craft something with the auto-crafting feature.

The problem is happening extremely infrequently and with no specific items. It's been happening to me for the last 3 days now. It wasn't doing it a whole lot at first, but now it's doing it for almost every craft... I have to retry up to 10 times before the pre-crafting GUI actually opens. 

I guess I could say it started happening when I started having a lot of items in my system. The more items I have, the more often it's happening (currently have around 30m items in it right now)

- Minecraft: 1.16.5
- Forge: 36.1.2
- Refined Storage: 1.9.12

#### Does this issue occur on a server? [yes/no]
No, singleplayer. Might happen on servers too though, haven't tested.

https://pastebin.com/UJj8XZe9

necauqua:
Fixes #2995

This thread was first added 4 years ago for item crafting preview and then blindly copied (I assume) for unfinished fluid crafting preview.

I don't know how the code structure was back then, but now this method call is enqueued onto the main minecraft thread from the network thread and does not really block anything.
Given that the calculation is also called on the main thread when updating nodes and it's totally fine, I think it's ok to remove this thread which causes concurrency issues (at least #2995). Or this can also be changed to call that calculation at the end of the world tick near all the node processing.

There are no other threads across RS and no code is threadsafe, you need to at least add a bunch of synchronization to ItemStackList/FluidStackList so that calling their `copy` from another thread (like it happens now) is safe and does not cause CMEs if you want to keep the preview calculation threads as they are now.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/handler/FluidGridHandler.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/handler/FluidGridHandler.java
@@ -99,40 +99,36 @@ public class FluidGridHandler implements IFluidGridHandler {
         FluidStack stack = network.getFluidStorageCache().getCraftablesList().get(id);
 
         if (stack != null) {
-            Thread calculationThread = new Thread(() -> {
-                ICalculationResult result = network.getCraftingManager().create(stack, quantity);
-                if (result == null) {
-                    return;
-                }
+            ICalculationResult result = network.getCraftingManager().create(stack, quantity);
+            if (result == null) {
+                return;
+            }
 
-                if (!result.isOk() && result.getType() != CalculationResultType.MISSING) {
-                    RS.NETWORK_HANDLER.sendTo(
-                        player,
-                        new GridCraftingPreviewResponseMessage(
-                            Collections.singletonList(new ErrorCraftingPreviewElement(result.getType(), result.getRecursedPattern() == null ? ItemStack.EMPTY : result.getRecursedPattern().getStack())),
-                            id,
-                            quantity,
-                            true
-                        )
-                    );
-                } else if (result.isOk() && noPreview) {
-                    network.getCraftingManager().start(result.getTask());
+            if (!result.isOk() && result.getType() != CalculationResultType.MISSING) {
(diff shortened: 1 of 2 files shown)


raoulvdberge:
Thanks!

raoulvdberge:
@necauqua  how where you able to reproduce this issue?

necauqua:
I actually didn't have it happen to me (yet), but from the stacktrace and my experience with CMEs (I had some) it was super-obvious what was was the issue.

I guess we need to ask the issue author if this fixes it for him, although I'm fairly certain it does.



Currently that separate thread calls ItemStackList.copy, which iterates through a (potentially pretty long) list, while that same list can be modified from the main thread by nodes or whatever, and CME is just thrown by the list exactly when this happens to prevent undefined behavior and/or hidden bugs.

And the solution, as I wrote, is either to ensure that the access to the list is exclusive (write a bunch of locks and whatever at the very least in ItemStackList) so that just never happens, or to remove the root cause - separate thread, what I ended up doing.

Darkere:
The chance for the CME is real, and the separate thread to prevent lag spikes is no longer needed as crafting calculation should be fast enough now. I think this is a good idea.

PR #3004, opened 2021-06-09
Allow addons to override crafter insertion

Darkere:
This will break the API for anyone implementing ICraftingPatternContainer. AFAIK there are 2 addons this will affect, I'll go and notify them if this gets merged. 

Just moved the insertion into the crafter along with some minor changes to reduce the amount of hackyness that has to happen for various usecases.


raoulvdberge:
Can you clarify why this is needed? What's the use case?

necauqua:
> This will break the API for anyone implementing ...

You can make those methods default and either define a default implementation of `insertItemsIntoInventory` as some util method somehow accessible through other API, or just move the current implementation to those interface default methods and that's it.

Darkere:
I can do some workarounds to make the API work without changes. However as this is not a thing anywhere else in the codebase I do not think this is something wanted.

necauqua:
What I described is not a workaround, even without aim for backwards-compatibility the hasConnectedInventory methods are just asking to be defaults

Darkere:
I guess so.

Witherking25:
I’m currently making an addon and this is perfect for what I’m doing

Witherking25:
btw crafters always say no machine found in this fork

Darkere:
whoops thank you

PR #3010, opened 2021-06-11
Fix pattern cache

### LINKED ISSUE #3009: Pattern left-click spreading in any inventory causes noticeable lag
Just a pattern item, unencoded, and a vanilla chest, right-click spreading (the one that does one item per slot) is totally fine, but left-click spreading (the one that splits items equally and does phantom rendering <- probable cause) actually lags the game.

Not like this is a huge deal but this is just weird and may hint at some pattern (rendering, I suppose) issue.

edit: observed in atm6 in a new world, just vanilla+RS does not have that, so maybe a number of recipes matters or something

I think it's related to pattern cache, it uses ItemStacks as keys, and they don't have equals/hashcode so it is essentially an IdentityHashMap (which is also never ever cleared it seems like).

edit2: yeah it looks like clientside often creates new itemstacks for rendering and the cache being an identity one keeps recalculating the same things, this is very visible when you open a big crafter interface filled full of patterns, full creative crafter causes freezes of up to a few seconds

PR is on the way

necauqua:
Current pattern cache implementation is very-very meh - it's never ever cleared, so it's a memory leak,
and minecraft ItemStacks do not implement equals/hashCode so the cache map works essentialy like IdentityHashMap, meaning that in certain scenarios performance is very bad.

Those scenarios include #3009 and also I was having an issue where opening big (e.g. 12*9 slots, the creative one) crafters filled with patterns would actually freeze the game for up to several seconds (continuing with fixes of ATM6 skyblock for myself).

Actually absolutely no clue how tf does it not crash (at least the client) with OutOfMemoryError, could've been spotted way earlier then.

This PR fixes both of them and if you think about it, should be an improvement of crafting performance

fixes #3009

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/tracker/ItemStorageTracker.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/storage/tracker/ItemStorageTracker.java
@@ -2,7 +2,7 @@ package com.refinedmods.refinedstorage.apiimpl.storage.tracker;
 
 import com.refinedmods.refinedstorage.api.storage.tracker.IStorageTracker;
 import com.refinedmods.refinedstorage.api.storage.tracker.StorageTrackerEntry;
-import com.refinedmods.refinedstorage.apiimpl.API;
+import com.refinedmods.refinedstorage.util.ItemStackKey;
 import com.refinedmods.refinedstorage.util.StackUtils;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ItemStack;
@@ -17,7 +17,7 @@ public class ItemStorageTracker implements IStorageTracker<ItemStack> {
     private static final String NBT_NAME = "Name";
     private static final String NBT_TIME = "Time";
 
-    private final Map<Key, StorageTrackerEntry> changes = new HashMap<>();
+    private final Map<ItemStackKey, StorageTrackerEntry> changes = new HashMap<>();
     private final Runnable listener;
 
     public ItemStorageTracker(Runnable listener) {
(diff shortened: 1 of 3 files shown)


Darkere:
For reusability it's probably better to use `API.instance().getItemStackHashcode()` and `API.instance().getComparer()` for the comparison. 

I believe there is even a similar ItemStackKey implementation being used somewhere else. Not sure if that could be reused. 

I don't think the memory leak is an issue. I mean it is technically possible to create thousands of patterns, but generally Modded servers have to be restarted (as in will crash due to a random bug) before ever getting close to that number. 

Do you have some profiling for this? I'm not certain the Cache is actually the problem (although it certainly isn't ideal).

tth05:
Did the same fix in my fork, because opening rebornstorage crafters and switching pages was extremely laggy. If the cache is fixed you will have lag when opening it the first time, but afterwards no performance issue.

necauqua:
> I mean it is technically possible to create thousands of patterns

The issue is that for example when the crafter is opened, it certainly creates a new set of item stacks that are never going to be freed, so if you open-close the crafter a bunch of times, the client is going to have leaking memory.

Yeah technically it's not really an issue, but it's an existing and proven source of a memory leak, hello

> I'm not certain the Cache is actually the problem

As soon as did this huge lagspikes when opening crafters completely are happening, just like @tth05 said, only the first time you open it.
I will actually ~~profile~~ sample it because I am curious if there is other source of that lag as well, but I doubt it.

tth05:
> The issue is that for example when the crafter is opened, it certainly creates a new set of item stacks that are never going to be freed, so if you open-close the crafter a bunch of times, the client is going to have leaking memory.

This shouldn't be an issue anymore with the new code though. Before it definitely was.

necauqua:
@tth05 actually yeah you are right, as I say in the comment above my cache reset, it should not actually happen, it's just a safeguard to never ever ever have a memory leak, in case maybe some mods do some weird stuff

necauqua:
image

yeah looks exactly like the spike I got

necauqua:
The main issue is in pattern rendering (the cause of crafter gui lag spike) and the fact that it does a whole bunch of useless things to extract the pattern result.

I think that you can write a separate method that just extracts the result from NBT (without touching e.g. the recipe registry) and the issue would be completely solved.

But well current cache is super-broken so this fix is needed anyway.

I'll try to do that later, sounds very doable

raoulvdberge:
Thanks!

raoulvdberge:
The cache was indeed kinda terrible, I implemented it as "good-enough", but never got around to fixing it.

PR #3011, opened 2021-06-12
Speed up pattern validation by using now-fixed cache

necauqua:
I wanted to make this commit a part of #3010 but didn't make it in time :)

Speed things up for vanilla-RS patterns by using that cache, I've never ever seen any addon making their own patterns so this is pretty much universal at this point.

Creating ICrafterPattern instances is quite expensive in modpacks with bagilions of recipes so while this isn't really an issue for vanilla RS it helps a lot in those modpacks.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/inventory/item/validator/PatternItemValidator.java
+++ b/src/main/java/com/refinedmods/refinedstorage/inventory/item/validator/PatternItemValidator.java
@@ -1,6 +1,8 @@
 package com.refinedmods.refinedstorage.inventory.item.validator;
 
+import com.refinedmods.refinedstorage.RSItems;
 import com.refinedmods.refinedstorage.api.autocrafting.ICraftingPatternProvider;
+import com.refinedmods.refinedstorage.item.PatternItem;
 import net.minecraft.item.ItemStack;
 import net.minecraft.world.World;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/refinedmods/refinedstorage/inventory/item/validator/PatternItemValidator.java
```diff
@@ -15,6 +17,13 @@ public PatternItemValidator(World world) {
 
     @Override
     public boolean test(ItemStack stack) {
+        return test(world, stack);
+    }
+
+    public static boolean test(World world, ItemStack stack) {
+        if (stack.getItem() == RSItems.PATTERN.get()) {
```
raoulvdberge, line 24: Why can't this check be done from the original `test` method?

necauqua: I first added the check and after that I made an impulsive refactor to move body of the test method to a static method, so that the whole check can be called from two other places without creating a throwaway instance, that's it.
I can revert the refactor part if you want

raoulvdberge: Please do

PR #3012, opened 2021-06-12
Curio support for wireless items + gradle updates

Darkere:
First 2 commits are just gradle, mappings and dependency updates (Used updateMappings command for updating mappings so those should all be exactly the same methods as before).

Was not able to fix curseforge api (I recall other people talking about it being broken) and switched to cursemaven. 

Also realized this kind of forces an addons update to add the tag for the wireless crafting grid.


Darkere:
At some point I'm gonna be a git pro and do things like you are supposed to. Today is not that day.

Close in favor of #3013

PR #3014, opened 2021-06-13
Implement ordered insertion

### LINKED ISSUE #1937: Insert order for processing patterns

necauqua:
This is the simplest and unlikely to be broken in any way implementation of #1937

Basically, I just changed `IoUtil.insertIntoInventory` from accepting unordered `Collection` to a `List`, and then went on from that.

This collection is obtained exclusively through ItemStackList.getStacks(), which is explicitly unordered as it merges stacks, makes sense.

So then I refactored a couple of methods that are used only in autocrafting to use lists of stacks instead of `ItemStackList`s, it was smooth as hell, and everything worked first try.

This means that those stacks are _not_ merged in a couple of cases, but I think that's actually desired, I can imagine a usecase for when the processing pattern can insert items in order, like A -> B -> C, and then A again into some specific machines.

I did some testing with ATM6, everything seems to just work.

resolves #1937

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
@@ -11,12 +11,17 @@ import com.refinedmods.refinedstorage.api.util.StackListEntry;
 import com.refinedmods.refinedstorage.apiimpl.API;
 import com.refinedmods.refinedstorage.apiimpl.autocrafting.task.v6.IoUtil;
 import com.refinedmods.refinedstorage.apiimpl.autocrafting.task.v6.SerializationUtil;
+import com.refinedmods.refinedstorage.apiimpl.util.FluidStackList;
+import com.refinedmods.refinedstorage.apiimpl.util.ItemStackList;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.world.item.ItemStack;
 import net.minecraftforge.fluids.FluidStack;
 
+import java.util.List;
+
 public class ProcessingNode extends Node {
+
     private static final String NBT_ITEMS_RECEIVED = "ItemsReceived";
     private static final String NBT_FLUIDS_RECEIVED = "FluidsReceived";
     private static final String NBT_SINGLE_ITEM_SET_TO_REQUIRE = "SingleItemSetToRequire";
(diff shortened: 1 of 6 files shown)


gurmiguel:
I've tested it on ATM6 merge into 1.9.15 and it seems to be working just as expected,
As @necauqua mentioned, this commit doesn't have any major changes, so it should present much risk.

Darkere:
@necauqua you got time to fix the merge conflicts?

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
```diff
@@ -249,8 +253,8 @@ public void updateFinishedQuantity() {
     public void onCalculationFinished() {
         super.onCalculationFinished();
 
-        this.singleItemSetToRequire = requirements.getSingleItemRequirementSet(true);
-        this.singleFluidSetToRequire = requirements.getSingleFluidRequirementSet(true);
+        this.singleItemSetToRequire = new ItemStackList(requirements.getSingleItemRequirementSet(true));
```
Darkere, line 256: Was there a reason to not just also convert `singleItemSetToRequire` into an Arraylist?

necauqua:
This is a quick and brainless rebase just makes everything compile and not crash, I didn't have a quick way to test this in 1.18 and I'm slightly busy currenly

So if this does not work because of some new mechanics I'll maybe do more after holidays :)

src/main/java/com/refinedmods/refinedstorage/apiimpl/autocrafting/task/v6/node/ProcessingNode.java
necauqua: No idea, don't remember :)

My guess is that those are returned in getters and those getters are used elsewhere and I wanted to keep the change minimal to increase the chance of it happening

Darkere:
oof. rebased to wrong branch. No worries I'll take care of it :D . Enjoy your holidays

PR #3060, opened 2021-08-07
 Fix override of Pattern color even when not in player inventory

### LINKED ISSUE #2980: Patterns change color but not icon when in grid and holding shift.
#### Issue description:
Patterns change color but not icon when in grid and holding shift.

#### What happens:
If you make a pattern with an output of a leaf or potion, place it into the storage grid (pattern grid specifically), it turns green for the leaves, and potion colored for the potions.

Either both the color and item model should be shown, or neither.

#### Steps to reproduce:
1. Make a processing recipe pattern with a single output of leaves.
2. Place it into the Grid (not crafter)
3. Hold shift with it in view
4. Pattern turns green

- Minecraft: 1.16.5
- Forge: 36.1.16
- Refined Storage: 1.9.12

Patterns In Grid:
Screenshot_20210516_194257
Patterns Outside Grid:
Screenshot_20210516_194240

TheJulianJES:
Fix #2980
This is the issue where RS Patterns with an overridden color (example: leaves) change their color (but not their model) in item frames, JEI favorites, an RS Grid, when dropped, and everywhere else that isn't the current players inventory.
​

The issue occurs because ``PatternItemColor.getColor(stack, tintIndex)`` only checks ``PatternBakedModel.canDisplayOutput(stack, pattern)``:
https://github.com/refinedmods/refinedstorage/blob/d98872c3101c8b2c2d2af21541771fe24ce7d0e7/src/main/java/com/refinedmods/refinedstorage/render/color/PatternItemColor.java#L15

The override for the actual model in ``PatternBakedModel`` -> ``getOverrideModel(model, stack, world, entity)`` first checks if the entity isn't null:
https://github.com/refinedmods/refinedstorage/blob/d98872c3101c8b2c2d2af21541771fe24ce7d0e7/src/main/java/com/refinedmods/refinedstorage/render/model/PatternBakedModel.java#L29-L32
​

To fix this and prevent further "edge-cases", a check is added in``PatternBakedModel.canDisplayOutput(stack, pattern)`` to see if the exact ItemStack (for the color override) is even in the players inventory.
(Note: ``Minecraft.getInstance().player.inventory.hasItemStack(stack)`` could not be used, as that just checks if the Item is present in the inventory at all (not the specific ItemStack))

(Tested in Singleplayer and Multiplayer)
​

Additional notes about the code:
- I'm not sure if the ``Minecraft.getInstance().player != null`` check is needed but it also doesn't hurt I guess.
- I would have liked to just check ``Minecraft.getInstance().player.inventory.allInventories`` (a combined list of all three inventory lists) but it's private.
- I also included ``armorInventory`` because this is also part of the players inventory and technically (either by bug or cheating), a pattern could be put there producing the same bug again.

If there's anything that could or should be done differently, please let me know and I'll update the PR.
(Maintainer edits are also allowed of course)

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/render/model/PatternBakedModel.java
+++ b/src/main/java/com/refinedmods/refinedstorage/render/model/PatternBakedModel.java
@@ -11,6 +11,7 @@ import net.minecraft.client.renderer.model.ItemOverride;
 import net.minecraft.client.renderer.model.ItemOverrideList;
 import net.minecraft.client.world.ClientWorld;
 import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.player.PlayerInventory;
 import net.minecraft.item.ItemStack;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


TheJulianJES:
I just noticed that this breaks the other two pattern render handlers in ``ClientSetup``:
https://github.com/refinedmods/refinedstorage/blob/d98872c3101c8b2c2d2af21541771fe24ce7d0e7/src/main/java/com/refinedmods/refinedstorage/setup/ClientSetup.java#L147-L176

I'll see if it's smarter to just implement in in the first handler there.

TheJulianJES:
I created another PR (with another approach to fix the issue): https://github.com/refinedmods/refinedstorage/pull/3061
I'll leave this one untouched.

PR #3062, opened 2021-08-07
Fix right-clicking NetworkItem in offhand locking wrong slot

### LINKED ISSUE #2997: Can't insert item from main hand when wireless or portable grid is opened in offhand
#### Issue description:
The portable and wireless grids understandably prevent you from inserting the grid itself into the storage system, but this is done by locking the main hand slot regardless of whether the grid is in it or not.

#### What happens:
Cannot shift-click or drag item from whichever hotbar slot happens to be selected, or drop an item in that slot if it was empty [shift-clicked items out of the storage system can land there though, oddly enough]

Slot behaves normally

#### Steps to reproduce:
1. Put the portable or wireless grid in your offhand
2. Open the grid
3. Try to use the slot that was selected for your main hand

- Minecraft: 1.16.5
- Forge: 36.1.13
- Refined Storage: 1.9.12

- Have not tried on a server

TheJulianJES:
Fix #2997

This is fixed by checking if the right-clicking action on the NetworkItem/wireless grid was performed by the main- or offhand.
If it's performed by the offhand, slot 40 instead of the currently selected slot on the hotbar "gets locked".
(This slot isn't actually visible in the GUI opened. Passing down ``-1`` also works in my testing but since ``40`` is apparently the slot used for the offhand, I used that.)

``Minecraft.getInstance().player.inventory.getStackInSlot(40)`` returns the ItemStack in the offhand. It doesn't look like there's a constant/value for this in Minecraft classes.

(``-106`` is apparently the slot used when saving the offhand as NBT(?))

Note: I also replaced `` player.getHeldItem(hand)`` in the same line with ``stack`` (as the value was initialized directly above and is the same).

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/item/NetworkItem.java
+++ b/src/main/java/com/refinedmods/refinedstorage/item/NetworkItem.java
@@ -38,7 +38,7 @@ public abstract class NetworkItem extends EnergyItem implements INetworkItemProv
         ItemStack stack = player.getHeldItem(hand);
 
         if (!world.isRemote) {
-            applyNetwork(world.getServer(), stack, n -> n.getNetworkItemManager().open(player, player.getHeldItem(hand), player.inventory.currentItem), err -> player.sendMessage(err, player.getUniqueID()));
+            applyNetwork(world.getServer(), stack, n -> n.getNetworkItemManager().open(player, stack, hand == Hand.OFF_HAND ? 40 : player.inventory.currentItem), err -> player.sendMessage(err, player.getUniqueID()));
         }
 
         return ActionResult.resultSuccess(stack);
--- a/src/main/java/com/refinedmods/refinedstorage/item/blockitem/PortableGridBlockItem.java
+++ b/src/main/java/com/refinedmods/refinedstorage/item/blockitem/PortableGridBlockItem.java
@@ -50,7 +50,7 @@ public class PortableGridBlockItem extends EnergyBlockItem {
         ItemStack stack = player.getHeldItem(hand);
 
         if (!world.isRemote) {
-            API.instance().getGridManager().openGrid(PortableGridGridFactory.ID, (ServerPlayerEntity) player, stack, player.inventory.currentItem);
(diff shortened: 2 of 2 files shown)


raoulvdberge:
I don't really like that you are using 40 since it can change any version.

raoulvdberge:
Does this work with the curios support as well?

TheJulianJES:
> Does this work with the curios support as well?

``onItemRightClick`` isn't called when the grid is opened through a keybind (which I think is the only way the grid can be opened when in a Curios slot). The only change that happens if the grid is right-clicked when in the offhand. (Even then, I didn't notice any issues with Curios items).

> I don't really like that you are using 40 since it can change any version.

Yeah. Although I think it's unlikely that it changes, it might be overseen when porting. 40 is only really used in ``PlayerInventory.storeItemStack`` (where the docs mention that the offhand is checked).
The value comes from ``allInventories`` (combining ``mainInventory``: 36 slots + ``armorInventory``: 4 slots + ``offHandInventory``: 1 slot. (Since starting at index 0, it's slot 40 for the offHandInventory)
Slot 40 is hardcoded again in ``PlayerContainer`` when adding the offhand slot. As far as I can see, there's no constant value in Minecraft classes.

I also noticed that the same thing needs to be implemented in ``PortableGridBlockItem``. (Tested and working)
Since slot 40 would be hardcoded twice for the offhand (in RS), is there some class where a constant for the offhand inventory slot could be put perhaps? (Would be noticed more easily when porting too.)

TheJulianJES:
Unrelated to this, it does appear that the Curios integration is somehow broken and experiences similar issues: https://github.com/refinedmods/refinedstorage/issues/3048
(Wireless crafting monitor not working when in a Curio slot and when a grid is used, it seems to lock the currently selected Hotbar slot.)

Since this PR isn't related to the Curios integration (but just to the offhand issue which seems to be fully fixed by this PR), I'd rather do a separate PR then (when I get around to seeing how the Curios API is implemented and if I get the locking issue fixed there).

Darkere:
Thank you this was fixed together with all the curio bugs in #3093

PR #3139, opened 2021-11-19
Add Refined Storage to Cardboard Blacklist & Moveable Blacklist

### LINKED ISSUE #3105: tag nodes non movable
Need to find the tag and put it on all the blocks. Prevents moving with pistons/contraptions + prevents picking up with cardboard box and similar.

starforcraft:
Closes https://github.com/refinedmods/refinedstorage/issues/3105


raoulvdberge:
This probably needs to be generated by a datagen

Darkere:
Well, it doesn't need to be. But it is nicer.
Also, adding mekanism tags is not the most ideal solution. 

Also also, there are tileentity tags that need to be done as well.

PR #3185, opened 2021-12-20
Use of requiresCorrectToolForDrops

Edivad99:
I wanted to take away the possibility of breaking blocks with your hands. 
Since minecraft 1.17 there is the possibility to add blocks to specific tags, `mineable/pickaxe` and `needs_iron_tools`.
Even if you do this, in the refined you still have the possibility to break the block with your hands🤔🤔.
I also updated the property of the rock, putting the second value in the strength method, the same as in minecraft


raoulvdberge:
We explicitly allow breaking with a tool. See `BlockListener#onHarvestCheck`.

PR #3277, opened 2022-04-11
fix integer overflow resulting in item lose

### LINKED ISSUE #36: Infinite fluid disk gets overflowed.
### Minecraft version

1.16.5

### Extra Disks version

1.16.4-1.5.1

### Forge version

36.2.22

### The latest.log file

(no related log found since i'm playing on server.) 

### Issue description

javaw_2022-02-11_08-09-29

I just moved from Blackhole Tank from Industrial foregoing(which can hold a fluid up to [max value of 32bit signed integer]mB) after I made Infinite Fluid Disk and got this.

### Steps to reproduce

Use Sink(from Cooking form blockheads) and Ultimate Importer(from Cable Tiers) to reproduce it faster. 

### Other information

I'm currently playing on All the mods 6 pack with version 1.8.17

If using disk manipulator with overflowed disk, It duplicates disk(but same id) and sometime it loses a data.

Also overflowed disk gets "red" signal on disk drive. 

Here's contents of disk:
image
Water goes `0` after it actually overflowed.
`Honey` from `Resourceful bees` almost near the max of 32 bit int.

MelanX:
This pull request fixes an issue with creative storage blocks/disks where an overflow of items results in vanished items/blocks. They are not shown in the grids anymore. My patches check for the space for a stack and only inserts until integer max value is reached. After this, no items of this type can be inserted anymore. 
Creating a new stack for this wouldn't help because it would still disappear in the grid.

*I just saw that I misnamed my branch for that. Ignore that, I mean `creative_disk_fix`.*

This PR fixes https://github.com/MelanX/ExtraDisks/issues/36


ManciDRaffy (dismissed):
works, tested it on my machine

Joshua154 (dismissed):
what a simple, working solution

raoulvdberge:
@MelanX  Can you add a changelog entry?

MelanX:
Done

raoulvdberge:
Can be a bit more descriptive. It's not a general item lose, only if you reach the maximum integer capacity (overflow)

MelanX:
I hope it's better now

raoulvdberge:
Thank you!

raoulvdberge (approved):

PR #3284, opened 2022-04-24
Fix shift crafting trusting storage cache too much

### LINKED ISSUE #3242: Make sure shift click extraction actually goes through
### Describe the bug

https://github.com/refinedmods/refinedstorage/blob/e3756e01265a749ed8de16e69c9e8b2b49454d19/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/CraftingGridBehavior.java#L122

If the extraction here fails, we are duping items.  Particularly problematic if people have slots with more than 64 items, as the default implementation of `ItemHandler#extractItem` specifies `Item.maxStackSize` as the maximum that can be extracted at once. 

### How can we reproduce this bug or crash?

Occultism dimensional matrix, with max stack size override enabled in config. 

### What Minecraft version is this happening on?

Minecraft 1.16.5

### What Forge version is this happening on?

36.2.22

### What Refined Storage version is this happening on?

any

### Relevant log output

_No response_

Darkere:
Fixes #3242. A fairly simple to use dupe bug. So it might be a good idea to backport this one.

This contains a slight behavior change for autocrafting. Extraction from inventories that use the default forge implementation, will now be much faster since the extraction will happen in the first tick. Shouldn't make much of a difference, though. 

Poll results: https://docs.google.com/forms/d/1K8F7qAZwh8Vo4lkunuORUVBOffI7bBKJhM596jVokXs/viewanalytics

The results of the poll confirm we are doing everything right. Although, we could consider changing this to produce 5 instead of 64. 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 - Improved Performance for grid updates
 
+### Fixed
+
+- Fixed a dupe bug related to shift crafting
+
 ## [v1.10.2] - 2022-03-26
 
 ### Changed
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/CraftingGridBehavior.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/grid/CraftingGridBehavior.java
@@ -13,7 +13,6 @@ import com.refinedmods.refinedstorage.apiimpl.network.node.GridNetworkNode;
 import com.refinedmods.refinedstorage.item.PatternItem;
 import net.minecraft.core.NonNullList;
 import net.minecraft.world.Containers;
-import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.inventory.CraftingContainer;
 import net.minecraft.world.item.Item;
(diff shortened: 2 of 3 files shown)


raoulvdberge (changes requested):
It is very difficult to understand `handleFailedCraft` and why everything in there is necessary. Please extract out some more methods, add some better comments, etc.

Darkere:
currently breaks simulation extraction. Need to fix before merge

raoulvdberge (changes requested):
Fix simulation.

Darkere:
After some thinking, I've come to the conclusion that the external storage extraction issue is super rare and in most cases probably already handled by the other side. 

Any solution from our side would be pretty ugly. And since autocrafting handles it just fine, it's a minor inconvenience if at all.

raoulvdberge:
So we don't merge this PR then? I'm not a very big fan of the fix.

Darkere:
I've removed the part of the PR, that fixes not being able to extract more than a stack through the default forge implementation (should have really been a separate PR).

The dupe fix is still there. 

A prettier solution would require a complete rewrite to make shift crafting no longer iterate over each craft. Lots of edge cases to discover there :/.

This code only applies if people break their storage cache and don't have enough items for shift crafting e.g. if they place 2 external storages on the same chest. 
Otherwise this does nothing.

raoulvdberge:

PR #3297, opened 2022-05-06
Change importer behaviour so they always try and import a full stack

TheOptik:
This PR changes the way the stack upgrade works inside of the importer along with some minor changes of the importers operation.
The changes mostly combat performance issues that arise when importing from quickly filling inventories.

Currently the Importer will import an `ItemStack` from a single inventory slot, no matter how many items are in that stack.

In the following configuration, two cables would be imported into the system every update cycle of the importer.

The proposed (oppinionated) change would cause the importer to take as many stacks as possible from the inventory, accumulate them into a single `ItemStack`, respecting the maximum stack size (and the `stackInteractCount`) as well as nbt data and then imports that single `ItemStack` into the system.
This results in a need for less importers (given a fixed item influx into the inventory) and less calls of the `insertItemTracked` method, which seems to be somewhat costly.

Secondly the proposed change increments the `currentSlot` variable every update cycle to prevent the importer from "getting stuck" on importing from a single inventory slot in an inventory that keeps filling up.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/node/ImporterNetworkNode.java
+++ b/src/main/java/com/refinedmods/refinedstorage/apiimpl/network/node/ImporterNetworkNode.java
@@ -4,6 +4,7 @@ import com.refinedmods.refinedstorage.RS;
 import com.refinedmods.refinedstorage.api.network.node.ICoverable;
 import com.refinedmods.refinedstorage.api.util.Action;
 import com.refinedmods.refinedstorage.api.util.IComparer;
+import com.refinedmods.refinedstorage.apiimpl.API;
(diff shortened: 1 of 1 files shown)


TheOptik:
After further testing of the code, i found that trying to import into a system that can not accept enough items causes item-loss.
Sorry for not realizing that sooner.
I will fix the code and update the PR after that.

raoulvdberge:
I'll close the PR for now, but I do like the change.

TheOptik:
Sorry it took me so long to get back to you.
Life kind of got in the way and if i am honest the PR slipped my mind a bit.
After seeing the notification i implemented the fix and pushed the code to my fork.
Would you like me to open a new PR with the changes or reopen this one?

raoulvdberge:
No need to say sorry, I know how it is! 
Feel free to open a new PR

PR #3320, opened 2022-06-08
LootItemFunctionType now use DeferredRegister

Edivad99:
From some versions Forge has added the ability to use LootItemFunctionType in DeferredRegisters.
This fixes an issue you will have when you upgrade to 1.19, it is also what they told me on discord to use in my mods.
I have set a minimal version which I am sure this change is available, this is why I have added some lines to the toml file

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -102,7 +102,7 @@ processResources {
 }
 
 dependencies {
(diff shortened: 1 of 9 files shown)


raoulvdberge (approved):

raoulvdberge:
Thanks!

Issue #2949, opened 2021-04-29
patterns lost using mekanism cardboard box on crafter


Publius2:
#### Issue description:
patterns lost when crafter moved using mekanism cardboard box. already tried reporting to mekanism and they said "Sounds like they don't write all their tile's data to NBT, which either they should or they should add it to one of the forge blacklist tags we look at so that it can't be picked up so this probably should be reported to them." this also happens with extra storage crafters but I don't know if you manage that add on
#### What happens:
crafter is empty when box is unpacked
#### What you expected to happen:
for the crafting recipes to be present
#### Steps to reproduce:
1. have crafter containing recipes
2.right click crafter with cardboard box
3.break box
4.place box in different location
5.shift right click box 
#### Version (make sure you are on the latest version before reporting):

- Minecraft:1.16
- Forge: Direwolf20 modpack 1.19
- Refined Storage:1.9.12

Does this issue occur on a server?
it might but I am playing single player and don't have a server to test on

#### If a (crash)log is relevant for this issue, link it here:

[pastebin/gist/etc link here]

Issue #2677, opened 2020-09-14
Invetory content deleted when using Mekanism cardboard boxes


csonyi:
#### Issue description:
When using Mekanism cardboard boxes on RS Crafters and Crafting grid, contained items (Patterns and Filters my case) disappear after placing them back down and taking off the box.
Found several issues about it on Mekanism's github page, all suggesting that the bug is in RS, not Mek, and that the reason is that the blocks mentioned above don't write their contents to NBT tags.

#### What happens:
RS blocks are placed, cardboard box placed over them, picked up, placed down elsewhere.
Inventory is wiped, and even though they are placed next to RS blocks that are connected to a system, they aren't powered most of the time (haven't tested it extensively, and I might be wrong about the " most of the time" part... When it happened, I moved ~12 crafters, so I made several trips, and I think a few of them did connect to the system, but their inventories remained empty).

#### What you expected to happen:
Items not disappearing.

#### Steps to reproduce:
Trivial.

#### Version (make sure you are on the latest version before reporting):
- Minecraft: 1.16.1
- Forge: 32.0.108
- Refined Storage: 1.9.2 (not the latest, but I didnt see a mention of anything relating to the bug in any of the releases since)

Does this issue occur on a server? [yes/no]
Happened in SP, didn't try in MP.

raoulvdberge:
Don't use block moving mods with RS, they aren't supported.

Issue #2439, opened 2020-02-25
[1.15.2] Mekanism Cardboard box and Cyclic Sack of Holding breaking stuff.


Quu1337:
#### Issue description:
Mekanism Cardboard box and Cyclic Sack of Holding can take secured/protected RS network blocks.
Both of them can rip Security Manager out of connected networks, cardboard box removes all Security Cards from it, making it useless.

#### Version (make sure you are on the latest version before reporting):

- Minecraft: 1.15.2
- Forge: 31.1.14
- Refined Storage: 1.8.1

Does this issue occur on a server? [yes/no] yes.

https://github.com/mekanism/Mekanism/issues/5899#issuecomment-590409035

"Due to how tags work I would say the modpack developer should be disabling things like the refined storage security manager. (Or refined storage themselves could just include things they don't think should be picked up in our tag, which would only make it do anything when both mods are loaded)."

https://github.com/Lothrazar/Cyclic/issues/1321

raoulvdberge:
Crazy. You'd think that RS blocks would immediately rip through cardboard...

Lothrazar:
Hey dev of cyclic here.

Should i check if its "unbreakable", or whats the best way to check "Can player harvest this"

raoulvdberge:
@Lothrazar  RS uses the BlockEvent.BreakEvent and cancels that if the block can't be broken.

PR #3321, opened 2022-06-11
Add datagen for Forge's block relocation blacklist, Packing Tape's block entity blacklist

### LINKED ISSUE #3105: tag nodes non movable

aeonic-dev:
This PR adds datagen for Forge's `forge:relocation_not_supported` block tag. The tag is used by Mekanism's cardboard box and other mods' methods for moving blocks, which inherently conflict with Refined Storage's data storage methods and result in lost/dangling data. As a side effect, we create colored block tags to mirror the item tags first, then add those and the remaining network node blocks to the blacklist tag.

We also add block entity tag datagen for [Packing Tape](https://www.curseforge.com/minecraft/mc-mods/packing-tape) (under `packingtape:blacklist/problematic`), another block moving mod that causes the same issue. 

There have been a few unresolved issues that were closed for various reasons (#2949, #2677, #2439) which this change addresses. Also closes #3105.


raoulvdberge (dismissed):
Awesome, thanks!

raoulvdberge:
@aeonic-dev Can you rebase on develop and add CHANGELOG entries?

aeonic-dev:
Yep, on it.

raoulvdberge (approved):

raoulvdberge:
Thanks!

Issue #3322, opened 2022-06-17
processing pattern creation crash


deadlywolf301:
### Describe the bug

was just trying to automate the making of the processers and when i attempted to make a processing pattern for smelting a raw processer the game crashed when i clicked the pluse button in jei to add the ingredints for the pattern.
[crash-2022-06-16_23.05.26-client.txt](https://github.com/refinedmods/refinedstorage/files/8924692/crash-2022-06-16_23.05.26-client.txt)

### How can we reproduce this bug or crash?

not shure because its an interaction between refinedstorage and jei but you could download my mod pack (refined exploration) to reproduce it exactly.

### What Minecraft version is this happening on?

Minecraft 1.18.2

### What Forge version is this happening on?

40.1.36

### What Refined Storage version is this happening on?

1.10.2.jar

### Relevant log output

```shell
java.lang.NoSuchMethodError: 'mezz.jei.api.gui.ingredient.IGuiFluidStackGroup mezz.jei.api.gui.IRecipeLayout.getFluidStacks()'
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.moveForProcessing(GridRecipeTransferHandler.java:177) ~[refinedstorage-1.10.2.jar%23278!/:1.10.2] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.moveItems(GridRecipeTransferHandler.java:152) ~[refinedstorage-1.10.2.jar%23278!/:1.10.2] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipeForPatternGrid(GridRecipeTransferHandler.java:95) ~[refinedstorage-1.10.2.jar%23278!/:1.10.2] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipe(GridRecipeTransferHandler.java:60) ~[refinedstorage-1.10.2.jar%23278!/:1.10.2] {re:classloading}
	at com.refinedmods.refinedstorage.integration.jei.GridRecipeTransferHandler.transferRecipe(GridRecipeTransferHandler.java:29) ~[refinedstorage-1.10.2.jar%23278!/:1.10.2] {re:classloading}
[...]

PR #3342, opened 2022-07-19
update to new JEI API

Darkere:
Currently still waiting on Mezz to backport this https://github.com/mezz/JustEnoughItems/issues/2901

This causes cover recipes to cycle regardless of what you've focussed. Not Ideal but not that much of a problem either. 

We may still want to wait since I don't think the JEI 10 has made it out of alpha for 1.18 yet.

### DIFF
--- a/src/main/resources/META-INF/mods.toml
+++ b/src/main/resources/META-INF/mods.toml
@@ -4,10 +4,10 @@ issueTrackerURL = "https://github.com/refinedmods/refinedstorage"
 license="MIT"
 [[mods]]
 modId = "refinedstorage"
-version = "123"
+version = "${version}"
 displayName = "Refined Storage"
-displayURL="https://refinedmods.com/refined-storage"
-logoFile="logo.png"
+displayURL = "https://refinedmods.com/refined-storage"
+logoFile = "logo.png"
 authors = "Refined Mods"
 description = '''
 An elegant solution to your hoarding problem
@@ -20,7 +20,7 @@ ordering = "NONE"
 side = "BOTH"
 [[dependencies.refinedstorage]]
 modId = "jei"
-mandatory = true
+mandatory = false
 versionRange = "[9.7.1,)"
-ordering = "NONE"
+ordering = "BEFORE"
 side = "BOTH"
--- a/src/main/resources/META-INF/mods.toml
+++ b/src/main/resources/META-INF/mods.toml
@@ -3,9 +3,9 @@ loaderVersion = "[40,)"
 issueTrackerURL = "https://github.com/refinedmods/refinedstorage"
 license="MIT"
 [[mods]]
-modId="refinedstorage"
-version="${version}"
-displayName="Refined Storage"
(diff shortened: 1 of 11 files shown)


mezz:
If you do this update, please specify a minimum required version of JEI 10.
If you want to have compatibility with JEI 9 and 10 together, you can target the latest JEI 9.7 and move away from deprecated methods in order to support both.

raoulvdberge:
Can we rely on JEI 9 (non deprecated methods) for 1.18 @Darkere ?

I think the API breakage between 10 and 9 is minimal if we move away from the deprecated methods.

I prefer to use a more stable JEI version for 1.18 long term.

Darkere:
Ah wasn't aware that was an option. For 1.18 that makes sense. Will look into it

mezz:
Looks good!
Please specify a minimum version of JEI 9.7.1 to ensure users know to install a version that supports the API features you are using:
https://github.com/mezz/JustEnoughItems/blob/6f97357d458f02f5cd57cc3dfe19dc674c265064/ForgeApi/src/main/java/mezz/jei/api/gui/ingredient/ICraftingGridHelper.java#L25

src/main/resources/META-INF/mods.toml
```diff
@@ -3,18 +3,24 @@ loaderVersion = "[40,)"
...
+modId = "forge"
+mandatory = true
+versionRange = "[40.1,)"
+ordering = "NONE"
+side = "BOTH"
+[[dependencies.refinedstorage]]
+modId = "jei"
+mandatory = true
+versionRange = "[9.7.1,)"
+ordering = "NONE"
```
mezz: I think you can make it so that JEI is not mandatory, but if it is present then it needs to be in the specific version range. If it is not mandatory, n ordering must be specified, but the specific ordering doesn't matter here so I just picked one.

```suggestion
modId = "jei"
mandatory = false
versionRange = "[9.7.1,)"
ordering = "BEFORE"
```

src/main/resources/META-INF/mods.toml
```diff
@@ -3,18 +3,24 @@ loaderVersion = "[40,)"
 issueTrackerURL = "https://github.com/refinedmods/refinedstorage"
 license="MIT"
 [[mods]]
-modId="refinedstorage"
-version="${version}"
-displayName="Refined Storage"
+modId = "refinedstorage"
+version = "123"
```
mezz:
```suggestion
version = "${version}"
```

mezz (changes requested):

raoulvdberge:
@Darkere  Can you add a changelog entry that states that we now require JEI 9.7 + remove the commented out lines in `build.gradle`?

raoulvdberge:
Fixes #3322

Darkere:
Figured I would leave them since we need them for 1.19. But I guess you can just copy them again.

raoulvdberge (approved):

raoulvdberge:
Thx!

PR #3347, opened 2022-07-31
Add Mineable with pickaxe to all blocks

gigabit101:
Added support to BlockTagGenerator.class


raoulvdberge:
Hrm, not sure I understand all these new block tags.

Anyway, I think I fixed it in bebeaad0ccb33ab072b725c56fd768649bf476e9 with fewer changes

gigabit101:
minecraft:mineable/pickaxe is set on every block in the game that can be harvested with a pickaxe, as seen here https://ss.gigabit101.net/javaw_f1SOMIFSe5.png

All the other commit is doing is lowering the default hardness of the block

raoulvdberge:
Yep, but lowering the default hardness makes the speed fine for breaking with + without pickaxe. So it's good I think.

Issue #3406, opened 2022-11-11
[1.19] Refined Storage lags when viewing certain recipes with JEI


SnakerBone:
### Describe the bug

When viewing recipes that use forge:wood, forge:planks, forge:stone, forge:glass, forge:potions, etc in JEI the whole game will lag and sometimes even crash. I believe its because it is trying to cycle through all the different variants. This seems to be a RS issue as AE does not have this problem

### How can we reproduce this bug or crash?

Try and remove recipe cycling for common tags (if possible) or perform extensive bug testing/fixing for this issue

### What Minecraft version is this happening on?

Minecraft 1.19.2

### What Forge version is this happening on?

43.1.47

### What Refined Storage version is this happening on?

1.11.1

Darkere:
Please run spark on the client while the lag occurs. So we can see what might be the problem.

SnakerBone:
Hello, here are the profiler results: https://spark.lucko.me/XzhPrcM47D 

When I started the profiler I viewed the recipe for the furnace (vanilla furnace) & the pity machine frame (from industrial foregoing) with the crafting grid open. Viewing the furnace recipe was a little laggy but viewing the pity machine frame recipe nearly crashed my game

Darkere:
You'll need to run sparkc instead of spark to profile the client side instead of server side.

(also just if you are interested your bees are lagging your game :D )

G33kman:
Hoping not to totally hijack this thread but I believe I'm having the same issue as OP.
Here's a link to my results: 

https://spark.lucko.me/EbGfygZBeE

I notice it's only when I go to view a recipe within RS if I open my inventory and use the search bar in JEI to do that same thing it shows everything without issue.

Hope this helps to see what's going on.

Thanks

Darkere:
Alright, that confirms my suspicion. Its the JEI resource tracker. Will need to optimize that at some point.

G33kman:
At least one of us knows where to look next. Lol
Appreciate the response!

ShroomAgent27:
Just want to add my profile too.

https://spark.lucko.me/nK1fjisjWv

Minecraft Version: 1.19.2
Forge Version: 43.1.47

AMT8 Version: 0.1.10

Using jars
refinedstorage-1.11.1.jar
refinedstorageaddons-0.9.0.jar
refinedcooking-3.0.2.jar
jei-1.19.2-forge-11.4.0.285.jar

raoulvdberge:
[...]

PR #3409, opened 2022-11-16
Attempts to Improve on Issue #3406 (Lag with JEI recipe with multiple possible ingredients)

Mg138:
*Note: this is not fully tested yet, but I don't see much of a possibility for it to break anything.*

Tested it both on a server(with the server's RS unchanged and without the fix) and singleplayer, both have significant performance boost.

As noted (somewhere) in GridRecipeTransferHandler, there's probably much better ways to fundamentally fix this.

For this is just a patch, please consider reworking the framework around Ingredient later in development, as I'm not able to provide such in a single PR or within a few hours of code reading.

However, for what it's worth, the patch is pretty effective on my machine.

*SIDENOTE: am I actually clearing the trackerCache correctly? idea is that it's supposed to be cleared when the GUI is closed, so it doesn't grow infinitely*

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/container/GridContainerMenu.java
+++ b/src/main/java/com/refinedmods/refinedstorage/container/GridContainerMenu.java
@@ -16,6 +16,7 @@ import com.refinedmods.refinedstorage.container.slot.grid.ResultCraftingGridSlot
 import com.refinedmods.refinedstorage.container.slot.legacy.LegacyBaseSlot;
 import com.refinedmods.refinedstorage.container.slot.legacy.LegacyDisabledSlot;
 import com.refinedmods.refinedstorage.container.slot.legacy.LegacyFilterSlot;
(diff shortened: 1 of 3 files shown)


Darkere:
When I made this I'm pretty sure it was only called once per recipe when it comes into view. Has this changed?

Rather than caching the tracker for every view, my plan was to make a Singleton tracker. Which would allow some more features like keeping track of incoming items to update the view more dynamically.

Mg138:
Honestly? I'm not too sure. Maybe I'll properly investigate into if this is actually something on JEI's end, but I should probably head to bed soon-ish.

Looking at it again, it doesn't seem that it should really have mattered, but maybe they did change it, meh.

If anything, that might just mean that the reallocation of that List<Ingredient> is much more expensive than anything else there.

Technici4n:
@Darkere It was changed to check every tick since the inventory's contents can change at any time. See mezz/JustEnoughItems#2951.

PR #3428, opened 2022-12-13
Rework JEI Ingredient tracking

Darkere:
This got a little big bigger than I expected it to be. But in addition to performing well it now also keeps track of incoming items while JEI is open.

It works by caching all items in storage in a map on the first tick. And on subsequent ticks we use that map + checking inventory and crafting matrix.


raoulvdberge:
Changelog entry please

Issue #3438, opened 2022-12-21
Game crashes when closing any Grid-Block GUI


schnow265:
### Describe the bug

I tried building a Network and testing it out. When I tried closing the Game, it hanged and crashed.

### How can we reproduce this bug or crash?

1. Build a Network
2. Place Grid-block
3. Open the GUI
4. Close using ESC

### What Minecraft version is this happening on?

Minecraft 1.19.2

### What Forge version is this happening on?

43.2.0

### What Refined Storage version is this happening on?

1.11.4

### Relevant log output

```shell
---- Minecraft Crash Report ----
// My bad.

Time: 2022-12-21 12:16:30
Description: keyPressed event handler

java.lang.NoClassDefFoundError: mezz/jei/api/ingredients/IIngredientType
	at com.refinedmods.refinedstorage.screen.grid.view.GridViewImpl.removed(GridViewImpl.java:88) ~[refinedstorage-1.11.4.jar%23169!/:1.11.4] {re:classloading}
	at com.refinedmods.refinedstorage.screen.grid.GridScreen.m_7861_(GridScreen.java:660) ~[refinedstorage-1.11.4.jar%23169!/:1.11.4] {re:classloading}
	at net.minecraft.client.Minecraft.m_91152_(Minecraft.java:966) ~[client-1.19.2-20220805.130853-srg.jar%23175!/:?] {re:mixin,pl:accesstransformer:B,pl:runtimedistcleaner:A,re:classloading,pl:accesstransformer:B,pl:mixin:APP:balm.mixins.json:MinecraftMixin,pl:mixin:A,pl:runtimedistcleaner:A}
	at net.minecraft.client.player.LocalPlayer.m_108763_(LocalPlayer.java:395) ~[client-1.19.2-20220805.130853-srg.jar%23175!/:?] {re:classloading,pl:accesstransformer:B,pl:runtimedistcleaner:A}
	at net.minecraft.client.player.LocalPlayer.m_6915_(LocalPlayer.java:390) ~[client-1.19.2-20220805.130853-srg.jar%23175!/:?] {re:classloading,pl:accesstransformer:B,pl:runtimedistcleaner:A}
[...]

PR #3446, opened 2023-01-02
Removed direct references to VanillaTypes.ITEM_STACK

eKKiM1:
Fix for issue #3438 

The constant value causes the JVM to try to link JEI classes at runtime when the callers class is loaded

Luckily JEI added convenience functions to remove the need for VanillaTypes.ITEM_STACK ( https://github.com/mezz/JustEnoughItems/commit/6a332a73f789565001677f5e814533d675d35fed )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 - Fixed some craftable items not showing as craftable in JEI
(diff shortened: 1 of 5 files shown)


Darkere:
~~Did you test this? I believe since the JEI imports are still there and being loaded it will still crash?~~

Was apparently tested. Not sure if avoiding calling the class completely if JEI is not loaded would be better.

raoulvdberge:
I don't think imports matter. It's the class reference that breaks. If this fixes it I'm all for it, it's pretty clean too.

Issue #3441, opened 2022-12-27
Grid search mode  “Normal autoselect” prevents JEI usage


rpop0:
### Describe the bug

Just as the title says, if using the “Normal autoselected” search box mode, you cannot use JEI as the grid search box takes priority over the input and you also can’t deselect the grid search box.

As a possible suggestion to fix this, clicking the JEI search bar could just deselect the grid search box.

### How can we reproduce this bug or crash?

1. Have refined storage and JEI installes
2. Set up a grid with items.
3. Set the grid search box mode to”Normal autoselect”.
4. Attempt to use the JEI search, it does not work.

### What Minecraft version is this happening on?

Minecraft 1.19.2

### What Forge version is this happening on?

1.19.2-4.3.1.55

### What Refined Storage version is this happening on?

1.11.4

eKKiM1:
I am unsure how the *autoselected* feature should behave.

Is the expected behavior:

1. that after each action the search box should get focus again?
2. that when the grids open the search box should be focused once. 

At this moment the implementations tends to try to achieve option 1 but is broken in multiple ways:

- JEI search box is not focusable because the RS Grid Searchbox tries to get focus after each action
- The forcing the focus breaks when some actions are performed (ex. double clicked on empty slot in grid)

This issue does not occur when using the *JEI syncronized - autoselected* mode because the input from the RS Search Box is syncronized to the JEI search box.

rpop0:
I believe the better functionality would be option 2, so that the search box gets autoselected only when opening the grid,because the current options renders JEI totally useless if you need to search with the grid open.

I personally dislike JEI syncronization because on multiple occasions I alternate between searching in JEI and in the grid.

Additionally, another mode could be added so there are two nods, one for each option you outlined above.

eKKiM1:
I am not a fan of adding another Grid Search Box Mode.

At this moment there are already 6 modes which i think are already too much.
For new users it's already hard to differentiate between them. Adding another one only increases the learning curve.

After some consideration i even think the amount of modes can be brought back to two:
1. Normal
2. JEI Syncronized

In normal mode the grid just focuses the Grid Search Box when opened (I don't know why we wouldn't want that by default?) and the Grid Search Box can lose focus (to interact with JEI)

In JEI Syncronized mode the Grid Search Box focuses on open and regains focus after each action. It's a non-issue we cannot select the JEI Search Box because the input to the Grid Search Box is replicated to the JEI Search Box (and thus synchronized)

[...]

PR #3448, opened 2023-01-05
WIP: Added a new alternative autofocus

rpop0:
In relation to issue #3441, the normal autoselect mode was unable to lose focus, thus preventing JEI usage when the normal autoselected mode was being used.

This PR implements a new search box mode, "Normal (autoselected alternative)", which allows the search box to be automatically focused on opening the grid, but it allows it to lose focus, thus enabling JEI usage.

Currently there is an issue with this implementation (The reason why this PR is WIP) that I am not quite sure how to fix because I am not familiar with the source code:

Using the "Normal (autoselected alternative)" mode, follow the steps:

- Search box starts autoselected, search for an item
- Click on the JEI search bar (This step does NOT defocus the search box, even though it should)
- Click anywhere outside the grid GUI (This step defocuses the search box)
- Click on the grid search box to re-focus
- Click on the JEI search bar (Now, the grid search box properly gets defocused)


raoulvdberge:
Sorry, I won't be merging new features for RS 1. If this needs to be changed we can discuss it for RS 2. Keep in mind that the search box settings have already changed slightly there.

PR #3464, opened 2023-02-07
losslessly opimize .png files

Jatie1:
All .png files losslessly optimized with [File Optimizer](https://sourceforge.net/projects/nikkhokkho/files/FileOptimizer/).

All images should be functionally identical due to the lossless compression plugins used.

Many assets got a 95%+ decrease in file size, unsure if this translates to less resource usage in-game but it can't hurt :)


raoulvdberge:
Hi unfortunately I don't accept PRs with (only) binary changes because I can't verify them properly. Thanks for the tip though, I'll try to apply this myself later.

PR #3531, opened 2023-07-05
Improve performance of insert and extract for Item Storage Disks

flamion:
Hi, we've had performance problems in combination with Reborn Storage due to the extremely large item disks. After profiling I've found that a significant amount of time (roughly 70 ms / tick) was spent on simply iterating through the items to count the stored amount. I've added a field to track it on insertion/extraction instead which improved performance massively in our case (lots of items, ~7500 insertions/extractions per second).
We've been running this modified version on our server for a few days now and had no issues.
I don't know if you want to merge it since it's for an old version, but I thought I'll just open the PR anyways.


raoulvdberge:
1.12 is no longer maintained. A similar fix has already been implemented for later versions.

Issue #3514, opened 2023-06-13
[1.19.2] Auto-Crafting Exception in server tick loop 


wiloxe:
### Describe the bug

Modpack: ATM8-1.0.18
Using Reborn Storage multi-block crafter.
When I added a new recipe to my RS system the server crashed almost instantly after.

I had an interface setup to request all slots to be filled with fluorite blocks from mekanism in stacks with crafting card, after I added the recipe for the fluorite blocks the server shortly ended up in a crash-loop.

I was unable to reproduce in singleplayer using simple setup. I am able to provide world save instead.

### How can we reproduce this bug or crash?

In my situation:
1. Interface with 1 stack upgrade, 1 crafting upgrade and 2 speed upgrades configured to fill all slots with 64 fluorite blocks connected to the system
2. Create the fluorite block pattern (Exact off, Not sure if that affects anything)
3. Put it in the multi-block crafter from Reborn Storage
4. Disconnected from server (Server crash)

### What Minecraft version is this happening on?

Minecraft 1.19.2

### What Forge version is this happening on?

43.2.11

### What Refined Storage version is this happening on?

1.11.6

### Relevant log output

```shell
https://pastebin.com/YTPMTGDH Latest.log
https://pastebin.com/pLFutmBS Crash log
```

raoulvdberge:
Hello. This is a bit late, but were you able to reproduce this issue easily in multiplayer? Like, constantly? Can you still provide us that world save?
I will close this issue until we know more.

raoulvdberge:
This bug has been fixed in v1.12.4 for Minecraft 1.20.1 and v1.11.7 for Minecraft 1.19.2.
Thank you for reporting.

raoulvdberge:
Hello, this bug is now also fixed for Minecraft 1.18.2 in v1.10.6.

PR #3589, opened 2023-11-05
fix: rare autocrafting crash

Darkere:
makes #3514 not crash anymore. The broken crafting task gets stuck instead


raoulvdberge:
Maybe add a warning log if this happens for further issues

Darkere:
that would massively spam the log if I did that

PR #3648, opened 2024-03-07
docs: add json file for Hungarian translation

GaalFlorian:
Hello again. I struggled with the rebasing of the commits, so the other pull request ended up closed. On the bright side, this one should no longer contain the absolute chaos that merge commits caused. I hope I finally got it right.


raoulvdberge:
Thank you for all the effort!

Issue #3132, opened 2021-11-16
Can't wrench behind a cover


raoulvdberge:
I can't wrench behind a cover.

RS 1.9.16

RS only

PR #3664, opened 2024-04-03
feat: wrench covers from the back

thakyZ:
Implements issue #3132

I did test it before committing. There are some issues though, if the sides of the cover is clicked it counts that as if you were clicking on the face that the side is on. Otherwise, works extremely well from whatever backend of the cover face you click as, so long as it's within a 2 pixel radius (given you are using a x16 texture pack)


raoulvdberge:
Thanks for the PR, but RS1 is no longer maintained and RS2 doesn't have covers.

PR #3669, opened 2024-04-05
Fixed dupe with External Storage and a higher stack size than 64

### LINKED ISSUE #3640: Dupe issue with mods that allow a higher stack size than 64
### Describe the bug

 If you attach an external storage to a sophisticated barrel/backpack/other  (container that can hold over 64 per slot) and then mass craft (shift right click output in a crafting grid) a craft that takes materials from there, it'll craft as many as the full amount can but only consume 1 stack per slot allowing a dupe glitch for compactable items and free items for other crafts.
 
 Example: 9 Stacks of diamonds in 1 slot, mass crafting diamond blocks will create a stack of diamond blocks but only consume a stack of diamonds from that slot, leaving you with a stack of diamond blocks and 8 stacks of diamonds.
 
 I've reported it to the sophisticated-mods dev and they explained that the issue is on RS's end and what it is.
 
 "Extract methods cap extracted amount to stack's maxsize, if you want to extract more than a stack you need to call extract multiple times"

### How can we reproduce this bug or crash?

1. Connect a sophisticated barrel/backpack/other to your system via external storage
2. Use stack upgrades to put over a stack of items in 1 slot
3. Use said item to mass craft something (shift right click on output in a crafting grid)

Said item will only have a stack of it consumed while creating an amount equivalent to how many it can make.

### What Minecraft version is this happening on?

Minecraft 1.18.2

### What Forge version is this happening on?

40.2.0

### What Refined Storage version is this happening on?

1.10.6

### Relevant log output

_No response_

starforcraft:
Fixes https://github.com/refinedmods/refinedstorage/issues/3640


Geekerandy:
Just curious here, is there any plans to backport this fix to Forge 1.19.2? I made a modpack that currently has this duping issue as described in #3640 .

mahjerion:
Hi, can this fix be merged? We're seeing this in Craft to Exile 2, version 1.20.1.

starforcraft:
Closing this as the PR can cause the game to freeze and RS2 is on it's way

PR #3715, opened 2024-07-17
tr_tr.json Turkish Localization [Can you add to the minecraft 1.20.1 version too ?]

RuyaSavascisi:


raoulvdberge:
Refined Storage 1 won't be updated for new translations, please contribute to Refined Storage 2 instead.

PR #3718, opened 2024-07-25
Fix poseStack translation leak in JEI RecipeTransferCraftingGridError

mezz:
Fix for
* https://github.com/mezz/JustEnoughItems/issues/3630

RS has a `poseStack.translate` with pushing and popping the stack, resulting in a leaked translation.

Before fix, hovering over the crafting button would shift GUI elements around:
Bug JEI - Refined

After fix:
Screenshot 2024-07-26 at 8 53 45 AM

I added mitigation for this in JEI as well but I think this is the "correct" place for the fix to be.

### DIFF
--- a/src/main/java/com/refinedmods/refinedstorage/integration/jei/RecipeTransferCraftingGridError.java
+++ b/src/main/java/com/refinedmods/refinedstorage/integration/jei/RecipeTransferCraftingGridError.java
@@ -1,5 +1,6 @@
 package com.refinedmods.refinedstorage.integration.jei;
 
+import com.mojang.blaze3d.vertex.PoseStack;
 import mezz.jei.api.gui.ingredient.IRecipeSlotsView;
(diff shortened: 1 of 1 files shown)


mezz:
Note this issue does not exist in `refinedstorage-jei-integration` for 1.21, the push and pop are there:
https://github.com/refinedmods/refinedstorage-jei-integration/blob/3ae49100628b59bee6c553e94f3ad4003e022755/refinedstorage-jei-integration-common/src/main/java/com/refinedmods/refinedstorage/jei/common/MissingItemRecipeTransferError.java#L49-L52

raoulvdberge:
Thx for the PR, but RS1 is no longer maintained. Fixed in RS2 like you said.

mezz:
Congrats on the RS2 release @raoulvdberge 👍

raoulvdberge:
Thank you <3

PR #3756, opened 2025-03-04
Update commands.adoc

itsjusttriz:
Fixes the 3rd column of the 1st table shown on the `/commands` docs page, as it is currently broken and values are placed in the wrong cells.


raoulvdberge:
Thanks, but with RS2 this change is no longer needed.

